PR #43, opened 2016-03-10
Fix processing of nested annotation type declarations

sviperll:
Allow meta-stuff to work with annotation type declarations defined as nested types.


phax:
I'm on vacation. Check it when I'm back :)

PR #49, opened 2016-07-07
Fix nested enums in meta

sviperll:


phax:
Is this enough for a new release?

sviperll:
I have no rush:) Personally I have no agenda for jcodemodel and plan no further changes in a mean time, so you can freely proceed with the release as you feel right.

phax:
I got quite quiet in the last time so I think people are happy with the "as is" situation (or they changed to some library???). Anyway I will release it to indicate progress :)

Issue #74, opened 2020-01-03
create resource directory


guiguilechat:
Since the check on packages name ( #70 ) , it's not possible anymore to create a resource in a resource dir that does not respect a package name.
Example, I can't create a new META-INF/data.txt file, because the META-INF does not respect the package name.

This was explained till https://github.com/phax/jcodemodel/issues/70#issuecomment-569739110 ; I create a separate issue for convenience.

The solution I think would be correct is
1. create the JResourceDirectory class (same level as JPackage)
2. move the method addResourceFile from JPackage to JResourceDirectory.
3. add sub-dir creation in this class, and the root JResourceDirectory in JCodemodel - basically like the package. Also the JCodemodel should have a Map of JResourceDirectory, by name.
4. add JPackage to JResourceDir translation, typically with ` public JResourceDir JPackage::asResourceDir(){return m_aOwner.getResourceDir(m_sName);}`
5. add again addResourceFile in JPackage , but calling corresponding resource dir 
`public AbstractJResourceFile JPackage::addResourceFile (@Nonnull final AbstractJResourceFile rsrc){return asResourceDir().addResourceFile(rsrc);}` ; and same for the other methods.

Possibly you can make JPackage extend JResourceDir. Still, overload the methods to call the resource dir when creating resources.

phax:
After my vacation next week :)

guiguilechat:
@fbaro

phax:
So, what I do to create a "META-INF/services" file, I do the following:
```java
    final String sContent = "bla";
      // Fake directory
      cm.rootPackage ()
        .addResourceFile (JTextFile.createFully ("META-INF/services/" + aEntry.getKey (), StandardCharsets.UTF_8, sContent));
```

guiguilechat:
should be 

cm.resourceDir("META-INF/services").addFile(aEntry.getKey (),  StandardCharsets.UTF_8, sContent) ; IMO.

phax:
Agree, would be nicer, but that is kind of a bigger change. The current solution is a valid work around, and I will leave this suggestion open.

guiguilechat:
you can make it without changing the major version (this is the reason of the item 5. )

phax:
@guiguilechat please check if that (3.3.1-SNAPSHOT) looks reasonable to you, or if there further improvements that make sense. Thx again for your input.

guiguilechat:
[...]

PR #75, opened 2020-01-20
clean JCFileNameHelper

guiguilechat:
- renamed unix-specific format with prefix UNIX_
 - renamed windows-specific format with prefix WINDOWS_ ; also removed _WINDOWS suffix
  - removed test on illegal chars for unix, since eg touch "?" or touch "<" are fine under linux (and result in the file being listed with ls ).

  see #74


phax:
Sorry, alreay done 🤷‍♂

PR #78, opened 2020-01-20
made tests use a unified test JCM - bad indentations

guiguilechat:
also reduces the overload of constructor.
see #74

### DIFF
--- a/src/main/java/com/helger/jcodemodel/AbstractJType.java
+++ b/src/main/java/com/helger/jcodemodel/AbstractJType.java
@@ -289,15 +289,16 @@ public abstract class AbstractJType implements IJGenerable, IJOwned
     return isAssignableFrom (aThat, true);
   }
 
-  protected boolean isAssignableFrom (@Nonnull final AbstractJType aThat,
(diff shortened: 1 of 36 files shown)


phax:
That is a huge change and I partially dislike it. It also breaks backwards compatibility.
First of all, please remove all the superflowous brackets - I hate them - thx.

Also please remove all the `serialVersionUID`  constants - that's bad style that they have the value 1 and they are created by the runtime automatically.

No public members in classes please.

For the other topics from #74 I will respond separately, but I don't have time for this today

guiguilechat:
Why exactly would it break backward compatibility ? I did not know following commits would be added to the PR

I don't know what are superflowous brackets ?
I have my coding style implemented in eclipse, so whenever I save a file it forces the code style on the lines I saved.  Do you have an auto format script so that my java files are automatically formatted to your own style when I push on git ? I already found one for the tab/spaces, seems like something possible.

I added the serialversion id because eclipse complains about them. I will add ignore to them instead  (in which classes ?)

I put members public in the exceptions because they are final and unmodifiable. But of course putting them as getX() is just as good.

PR #81, opened 2020-01-23
eclipse formatter + cleanup actions, update readme with eclipse settings

guiguilechat:


guiguilechat:
formatter was only tested on JCodeModel class.

PR #83, opened 2020-01-23
[WiP] preprocessors

guiguilechat:


phax:
Is that branch still WiP?

glelouet:
Again, I was not sure if the coding style was satisfying. The PR was not supposed to be used as is, but rather give you ideas of how to do it in that specific case.

glelouet:
However there was some issues with that branch so I advise against using it without giving it some thoughts.

Issue #94, opened 2021-01-22
[feature] copying JCodeModel


glelouet:
I propose to add a new method in JCodeModel : copy().
This method returns a new CopiedModel that contains recursive copy of all the elements of the parent model.
This CopiedModel also allows to translate the elements of the parent model into its own elements.

This method could be used with the preprocessors : instead of working on the original model, processors would be applied on the copy. If no processor is used, then no copy is done.

Working on it.

phax:
That means a deep-copy of all objects and expressions... That's going to be a big one

glelouet:
yes.

phax:
Okay, than can we have quick chat on how to do it?
I don't like `Object.clone` because you always need to cast.
Can you do something with copy constructors? 
Or do your prefer a custom clone like in https://github.com/phax/ph-commons/blob/master/ph-commons/src/main/java/com/helger/commons/lang/ICloneable.java ?

glelouet:
I'll make a branch.

I was more asking if there is a visible impossibility for this, as to avoid doing "a big one" for nothing.

The method in JCodemodel is just copy(), the ModelCopy extends JCodemodel and add various translate(..) to get the translation.

phax:
One thing that strikes me, is that the basic types (like VOID or BOOLEAN) depend on the JCodeModel instance and all usages of those must also be transfered. That means that the new CM must be passed through quite deep

glelouet:
The issue I have, is that some constructor eg Jatom() are protected.

phax:
Just make them public

glelouet:
NAh I put my ModelCopy in the same package.

The less modifications, the better.

glelouet:
https://github.com/guiguilechat/jcodemodel/blob/copying/src/main/java/com/helger/jcodemodel/ModelCopy.java

phax:
I though you add a copy-constructor or clone method in every class?
Do I get you right - you manually copy the structure from the "outside" of the relevant class?

glelouet:
yes, I just add one class that makes the copy in itself.

phax:
Ah okay. But isnt't there a certain danger that you forget stuff?

glelouet:
indeed.

phax:
Than maybe for the expression stuff could you go for a solution like this: https://github.com/phax/jcodemodel/commit/ead1645d882ff301a530480c26e1e6d7b2c1348d

glelouet:
[...]

PR #96, opened 2021-02-17
Copying

glelouet:
#94 

some changes are formater-related.

basically add a copy()  method in the JCM to copy itself. Also a test class for this new method. plus a few fields that are given accessors, and classes that are made implement serializable.


glelouet:
@phax I think the correction is not good, throw exception without a message would be better.

```java
throw new UnsupportedOperationException (e);
```

PR #107, opened 2025-08-22
Add scripts

glelouet:
I added scripts to a new sh/ dir

1. `sh/install` runs maven install. This allows to have specific parameters for dev env, typically dev-specific profiles enabled.
2. `sh/upgrades` lists the possible upgrades to plugins and dependencies. It does NOT make any change to the actual project, only tells you some things exist.
3. ` sh/spacetabs` forces the git configuration that was explained in the readme, in order to commit java files with doublespace.

mvn params I use in those scripts : 
 - `-ntp` prevents flooding the logs with transfers when a server is slow/down. I got a project where the transfer progress was too verbose and above the logs limit, making the CI build unreadable.
 - `-T2C` tells maven to use twice the number of cores available for parallel modules building in the reactor. Especially important when lots of small submodules and you deploy them (so mix of local IO, remote IO and CPU usage)

Maybe later I will add a `sh/cores` that only install the core modules (not the examples, not the plugin) typically useful when adding a new feature on core and not willing to wait yet for full tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,9 +181,12 @@ Pull requests must follow my personal [Coding Styleguide](https://github.com/pha
 
 ### Tabs vs spaces
 
-This project uses double-space for indentation. If you want to use tabs, you can ask git to modify the files when commiting them and when pulling them. For this, from your project directory : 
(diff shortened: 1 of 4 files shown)


glelouet:
@phax  too fast !  :D

PR #108, opened 2025-08-22
moved versions in root pom and added last plugin version for some + c…

glelouet:
1. move one version from the main jcodemodel to the parent
2. added several maven plugins managements, most only for version, but a few with config. Each plugin has its mvnrepository url specified for fast version change check. This is important to be sure an error on a location does not happen because it uses a different plugin version.

plugin configurations : 
 - maven-source-plugin has jar-no-fork. I think it's already called. I had issues when not using it in multi-modules projects
 - versions-maven-plugin ignores the various `-RCX` versions as well as 1.2.3.4 versions.
 - The maven-compiler-plugin is not configured because I think it's not up to me.

----
 
 the configuration of the maven-compiler-plugin that I would do is
```xml

					<configuration>
						<release>${java.version}</release>
					</configuration>
```
But this also needs to add a property

```xml
	<properties>
		<java.version>17</java.version> 
```
Which is not mine to decide.

----

The versions plugin complains about maven version not set. Ignore it.

FWIW here is the result of `sh/upgrades`

> [INFO] --- versions:2.18.0:display-dependency-updates (default-cli) @ jcodemodel-parent-pom ---
> [INFO] The following dependencies in Dependency Management have newer versions:
> [INFO]   com.fasterxml.woodstox:woodstox-core .................. 7.0.0 -> 7.1.1
> [INFO]   com.sun.istack:istack-commons-runtime ................. 4.1.2 -> 4.2.0
> [INFO]   com.sun.istack:istack-commons-tools ................... 4.1.2 -> 4.2.0
> [INFO]   jakarta.servlet:jakarta.servlet-api ................... 5.0.0 -> 6.1.0
> [INFO]   jakarta.servlet.jsp:jakarta.servlet.jsp-api ........... 3.0.0 -> 4.0.0
> [INFO]   org.apache.ant:ant ................................ 1.10.14 -> 1.10.15
> [INFO]   org.glassfish.external:management-api ................. 3.2.3 -> 3.3.0
> [INFO]   org.glassfish.gmbal:gmbal ............................. 4.0.3 -> 4.1.0
> [INFO]   org.glassfish.gmbal:gmbal-api-only .................... 4.0.3 -> 4.1.0

### DIFF
--- a/jcodemodel/pom.xml
+++ b/jcodemodel/pom.xml
@@ -33,7 +33,6 @@
     <dependency>
       <groupId>com.github.javaparser</groupId>
       <artifactId>javaparser-core</artifactId>
-      <version>3.27.0</version>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -19,7 +19,7 @@ License file that accompanied this code. Modifications: If applicable, add the f
 	<description>Java code generation library parent POM</description>
 	<url>https://github.com/phax/jcodemodel</url>
 	<inceptionYear>2013</inceptionYear>
-	
+
 	<licenses>
 		<license>
 			<name>CDDL+GPL 1.1</name>
@@ -91,7 +91,7 @@ License file that accompanied this code. Modifications: If applicable, add the f
 			<url>https://github.com/sviperll</url>
 		</contributor>
 	</contributors>
-	
+
 	<modules>
 		<module>jcodemodel</module>
 	</modules>
@@ -106,6 +106,14 @@ License file that accompanied this code. Modifications: If applicable, add the f
 				<type>pom</type>
 				<scope>import</scope>
 			</dependency>
+			<!-- sub modules-->
+			<dependency>
+				<groupId>com.helger</groupId>
+				<artifactId>jcodemodel</artifactId>
+				<version>${project.version}</version>
+			</dependency>
+
+			<!-- other dependencies-->
 			<dependency>
 				<groupId>org.eclipse.platform</groupId>
 				<artifactId>org.eclipse.equinox.common</artifactId>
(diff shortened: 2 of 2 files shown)


phax:
You don't need it declared in your pom.xml to run it explicitly.
You may also consider a version exclusion list as e.g. https://github.com/phax/meta/blob/master/deps/versions-maven-plugin-rules.xml

phax:
And please don't bloat he parent POM with all kind of specific versions. If you want a Maven plugin as submodule, please define the specific stuff there.

phax:
This artefact is not needed anymore. I think they call it now Maven Model or so. See e.g. https://github.com/phax/ph-css/blob/master/ph-csscompress-maven-plugin/pom.xml for the artifacts needed for a Maven plugin.

And I absolute dislike having tooling specific dependencies in the parent pom. It's fine for me if you define a version property in there, but don't include the dependency - been there, and it's not ideal when working in complex project setups. Thanks

phax:
Yes I understand, but they have an impact in other POMs when the parent POM is included as a BOM.
See e.g. SpringBoot - when I include this in my POM:
```xml
  <dependencyManagement>
    <dependencies>
      <dependency>
        <groupId>org.springframework.boot</groupId>
        <artifactId>spring-boot-dependencies</artifactId>
        <version>3.5.4</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
    </dependencies>
  </dependencyManagement>
```
I get a predefined version for some PostGresSQL JAR - and this is totally annoying and confusing. So I am also trying to limit the dependencies in the `<dependencyManagement>` section of a BOM to only the ones absolutely needed for the core functionality.
Hope that makes sense to you.

glelouet:
I moved all versions back in the main module and hardcoded the versions plugin call in the script. we can discuss the changes later.

phax:
Thanks. Let me know when the PR is ready

glelouet:
please check that the deps are indeed needed in the main project. I only had two errors when I force merged your version ^^

PR #109, opened 2025-08-22
Plugins

glelouet:
several things added to test the plugins and a minimal generator

1. jcodemodel-maven-plugin is added. This plugin needs to be given a generator dependency that it will use to build a JCodeModel.
2. helloworld generator is added. This generator generates a single class with a single String field, default Hello::value="world".
3. helloworld-example is added. This module uses the maven plugin, with its dependency helloworld , to generate that single class. The parameters passed to the plugin are changed in a second execution, with different dest folder, way to specify the generator class, and params to generate the class. A JUnit test validates that those classes exist, and have correct value field.

The src/generated/java dir is created by the plugin, while this path is used by the compiler plugin and clean plugin. Same for src/generated/java2/ (for second execution).

### DIFF
--- a/plugin/plugin/pom.xml
+++ b/plugin/plugin/pom.xml
@@ -19,6 +19,7 @@
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-core</artifactId>
       <version>3.9.11</version>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.apache.maven</groupId>


glelouet:
There are two ways to specify the generator : 
1. if the plugin is configured for a specific class it loads that class.
2. otherwise the plugin looks for a  jcodemodel/plugin/generator file containing the class to load, typically each generator should have it (I should add a processor for that to be made automatically but that will add more modules and it's not needed - yet)

So as long as one generator is specified in dependency of the plugin, it will be loaded and the plugin will have access to the file.
The example in the second execution hardcodes that call. Hmm a change to add.

phax:
Someone's motivated ;-)

glelouet:
Note that I did not add a clean/rm feature on the mojo because this is already present with the maven-clean-plugin.  Less doublon feature = less bugs.

PR #113, opened 2025-09-08
Json and yaml generators

glelouet:
Depends on https://github.com/phax/jcodemodel/pull/110 which adds the flat structure management used in those modules.

The two generators are very similar. In fact, the yaml plugin extends the json plugin with only a change of parsing using jackson yaml  factory ^^ 

The tests are made only in the yaml one because it's easier to write them.
The json example still creates structure, but it's very basic.

The yaml test source is present as a file instead of as the pom's "data" configuration because the pom tabs can be changed by git and tabs are not allowed in yaml.

### DIFF
--- a/examples/plugins/yaml/pom.xml
+++ b/examples/plugins/yaml/pom.xml
@@ -34,6 +34,7 @@
             </goals>
             <configuration>
               <source>sourceConcrete.yaml</source>
+              <rootPackage>com.helger.jcodemodel.examples.plugin.yaml.concrete</rootPackage>
               <params>
                 <concrete.list>LinkedList</concrete.list>
                 <concrete.map>LinkedHashMap</concrete.map>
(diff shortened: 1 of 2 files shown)


glelouet:
@phax I think its good. 
This adds more feature than expected. Did not want ot make new PR , but I will stop with that one now.

phax:
Okay thanks - I am currently swamped with other stuff, but I intend to review it asap - thanks for your hard work - really appreciate it

glelouet:
@phax concrete type spec is added.

glelouet:
I added a rootpackage option.
This option is present at plugin level and applied to the generator.

```xml
            <goals> <goal>generate-source</goal> </goals>
            <configuration>
              <rootPackage>com.helger.jcodemodel.examples.plugin.csv.resolve</rootPackage>
              <data>
              …
```

All classes *created* are prefixed this root package if needed.

Some examples use it, some use it partially(they specify subpackage classes), some don't (full package specification).  
The local names are removed a leading "." to allow root package specification in csv ("p ." specification) . Otherwise empty package name would be trimmed and that would create the class "p" :P 

This feature reduces the verbosity of the csv and json examples:

https://github.com/phax/jcodemodel/pull/113/commits/e448721493206ea26ba029dbdd66ae33cf9d3678#diff-36c2145ded573e7def36f76c8a3e0c5292a3ff19bb0dd7e6519e98ed3678d9fc

This also means we can produce data from the same source file, with different plugin options, in different package.

PR #118, opened 2025-09-24
apply redirct process methods by order

glelouet:
redirected methods are now sorted by:
 1. method name
 2. number of arguments
 3. method signature (using Method::toString)


glelouet:
This is taken from 
https://github.com/phax/jcodemodel/pull/116/commits/0497cf98f55976e7364b4a0bad5285af9eafd821#diff-ac0458e421d349f3f27d56b664f318acdb7775c70c90e9d07b2b8ddfefda6e6dL572
Since I'm not sure I will keep that branch.

The constraint I comply to, is having the processing deterministic. 
The internal JdefinedClass methos are stored in a linkedhashmap, however the java.lang.reflect methods are not, so this needs to be sorted.

glelouet:
should be good regarding autosave actions (I did not add the file after autosave were triggered :evilcat: )

PR #128, opened 2026-02-03
Records

phax:
Based on #126


phax:
Syntactic sugar with fixed coding styleguide for getters and setters ;-)

glelouet:
> Syntactic sugar with fixed coding styleguide for getters and setters ;-)

Yeah I like that refthis can take a record into account. Something required when doing, say, record constructors :P

phax:
Will do it later - have other priorities atm

Issue #129, opened 2026-02-03
Test module ?


glelouet:
The test in JCM mostly use code matching.

This is weird because this should instead be checked using compiler.

I propose the following changes : 

1. a /jcodemodeltests module, with whatever groupId:artifactid ; depends on main JCodemodel module.
2. a `@JCMTested` annotation to place on the test classes that generate JCodemodel instances
3. an annotation processor that generates the corresponding classes in src/test/generated 

This way I could have a test class in src/main/java with 

```java
public JCodemodel createSimpleIntRecord(){
// do the creation and return the CM
}
```

and the corresponding test in src/test/java with 

```java
  @Test public void testSimpleIntRecord(){
  SimpleIntrecord sr = new SimpleIntRecord(24);
}
```

That's the basic idea, if yo have better one I'm all hear.

glelouet:
Can't work with annotationprocessor because it can't give access to the Class the annotation is applied to - therefore I can't call the methods.

glelouet:
To be fair, that's the behavior I was having in other CPP projects about 20 years ago, regarding auto generating drivers based on their specifications.

glelouet:
I used a plain file scanner in https://github.com/glelouet/jcodemodel/blob/%23129_testmodule/jcodemodeltests/src/main/java/com/helger/jcodemodel/compile/annotation/GenerateTestFiles.java

This lists all the classes in the src/main/java, loads them with classloader, validates annotation is present, then lists the public methods that produce a JCodeModel without params ; for each of those, invoke it and apply the writer on the generated JCM.

The class is executed by the exec-maven-plugin at generate-test-sources phase, so before test classes are actually compiled.

glelouet:
@phax  I can't PR in a new branch here. Can you add new branch "#129 _testmodule" please ?

glelouet:
No check, I close and let you see with the PR. No need for two discussion places.

PR #131, opened 2026-02-27
#129 testmodule

glelouet:
Have not worked it in a few weeks so I forgot details but …

Implementation for #129

1. add a module dedicated to simple tests : JCodeModel-Tests in /jcodemodeltests
2. generating classes are in the src/main/java ; annotated with new annotation `@TestJCM`
3. their public method that produce a JCodemodel from nothing are called after instantiation (or statically if static) 
4. the JCM is then written in the src/generated/javatest which is bound to test phases
5. the actual tests using those classes are in the usual src/test/java.

This mean you can see what is actually created, work with generated behaviour (and not code only) since you can load the generated class in tests.

 I recommend using the same method names for the generation and the test so as to know what you are testing exactly ; and suffix "Gen" at the end of the class generating (the one that is annotated `@TestJCM`) .
 For example, I want to test the record, RecordTestGen is annotated in src/main/java, has a simplerecordtest() method ; and RecordTest is the `@Test` in src/test/java, has the same simplerecordtest() method.
 
 I only integrated the records tests


glelouet:
@phax  opinion on the feature ?
I guess barely useful without explanation.
IMO there are several benefits :

1.visually see the result as we make the tests
2. load the generated test classes in tests for complex unit tests
3. keep an history of what is changed in the generation after a change : we know *what* trigggered a change say in indentation.

phax:
Well it was more tests and examples - tests and examples are always good. It was a pragmatic approach without investing too much review time - your contributions usually make sense :-)

PR #134, opened 2026-03-01
Bom

glelouet:
Note : this PR depends on #131 .

This PR integrates the incoming notion of POM-BOM in maven 4.
https://maven.apache.org/whatsnewinmaven4.html

1. A "bom" project (submodule) is created below root.
2. This bom project, type pom, only specifies dependencies and plugins management that are provided by the (jcodemodel) project. Once maven 4 is released, we can just change its type from POM to BOM. 
3. each other project below root has its parent changed from root-pom to bom, relativePath also changed.
4. The deps/plugin managements specified in root-pom are removed (moved to the bom)

Usage : 

Most user should not have anything to do, they will keep importing jcodemodel . For the ones who need several several of the proposed projects, they should import the BOM, as explained in the readme


phax:
In my other projects, the parent POM is directly my BOM. What is the idea of separating those two?

glelouet:
I forgot to link the doc …

glelouet:
https://maven.apache.org/whatsnewinmaven4.html#New_packaging_type.3A_bom

phax:
I think there are two things break the BOM contract:

1. `<pluginManagement>` only propagates through `<parent>`, not through `<scope>import</scope>`. The block for `jcodemodel-maven-plugin` therefore does nothing for external consumers of the BOM — it should stay in `jcodemodel-parent-pom`.
2. Using the BOM as the `<parent>` of every internal module conflates the BOM role (published, dependency-only) with the parent-POM role (non-published, build/plugin config). Please keep the modules' parent as `jcodemodel-parent-pom` and let the BOM be a leaf with `<dependencyManagement>` only.

Minors:
* rename the artifactId to `jcodemodel-bom` (matches `jcodemodel-parent-pom`, `jcodemodel-maven-plugin`)
* add a trailing newline to `bom/pom.xml`

glelouet:
hoo that sucks. I want to be able to specify the plugin versions too as bom :'( So people could just use the plugin and TADAAAAA correct plugin version and correct plugin dependencies.

Yes in the maven 4 we won't need to specify the version of the siblings anymore. The parent=bom is here to avoid duplicating spec ( I hate data duplication :P ) so will be useless in maven 4.

Remember this is to prepare maven 4 :)

Also maybe I misunderstood your previous question. If that's the case, this is not a hurry we can wait a bit for specs to be stable (audible sound horse) .

phax:
Yes okay - so we leave this one open for now

PR #139, opened 2026-05-09
generated/javatests licenses

glelouet:
Small fix to add the licence so the generated class files.
Copied from the JCMWriter to dicover the classes from the packages

 - lists the classes of the packages
 - hidden classes are discarded
 - classes with already a header are discarded (even if empty header)
 - subclasses are not walked over.
 - add the header LICENCE to the selected classes.

@phax  I let you change what needs change.
maybe run sh/cleaninstal before merging ?


glelouet:
I was thinking of extracting the classes iteration in a dedicated JCM proxy field
something like `Stream<JDefinedClass> JCM#classes::rootClasses` , with `classes` the proxy field (basically a structure to contain existing classes listing etc)

phax:
Thanks for your efforts @glelouet !!

glelouet:
I told you in another one that this was easy :)

I know the pain to have to read through dozens of commits and trying to decipher what the heck the author was trying to do, just to realize that this was removed later on.

PR #143, opened 2026-05-22
Transmit Java Feature

glelouet:
# What is java feature

It's the major part of the installed java version, for example 127 for java 127.0.0.1 .

Allowed java features are the same among different versions of a same java feature, and typically kept later.
For example lambdas were added in java 8 . So the lambdas features requires java feature 8+. A code that uses the "lambda" feature will fail to compile by javac <8 ; and the class files generated by javac from this code will fail to run on a jre <8 .

This value  is stored inside the java .class files produced by javac.

# What is Java release

It's the same, but at the scope of a project. The java release, also known as java.version, is set in the maven project to know what version to validate the java source files, as well as what feature to store in the compiled .class files.

The java installation used to build with maven must be at least this feature. If I have jdk 7 I can't build project with release 8.

Any feature from a later java release is disabled and therefore their inclusion in the project will make the build fail.

This is important because in the event we work with a more recent java installation, we still want the java generated class files to be runnable on an older java version. 

# Why it's needed

If we want to incorporate recent features, those need a way to generate replacement code when the target java release is below the required java feature of that feature.

For example, If I want to add switch patterns, then I need java feature 25, but if the java release is  lower, I need to use fallback code instead (so a series of instanceof at the top of the default case).

For this, the JCM code must be aware of that value. This is what this PR provides.

# Changes

1. default java feature added in the JCMWriter so any later change can be set here.
2. added JCMWriter.javaFeature property. 
3. added JFormatter.javaFeature property. This is passed from JCMWriter
4. passing the java.version to the system property java.feature in the jcodemodeltests, using it in the GenerateTestFiles
5. adding a plugin property jcodemodel.java.feature to transmit it in the plugins

That's it. Nothing else changes, since nothing yet uses it. but now the feature value is visible, modifiable, the default value can be changed easily.

### DIFF
--- a/jcodemodel/src/main/java/com/helger/jcodemodel/JExpr.java
+++ b/jcodemodel/src/main/java/com/helger/jcodemodel/JExpr.java
@@ -40,6 +40,8 @@
  */
 package com.helger.jcodemodel;
 
+import java.util.stream.Stream;
+
 import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.Nullable;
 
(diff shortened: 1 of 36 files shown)


glelouet:
not compiling anymore after merging the last master.

glelouet:
I give up. 30 min to merge and it still does not work.

phax:
Let me see what I can do

phax:
Implemented directly in master

PR #144, opened 2026-05-22
Switchexpression for java 17

glelouet:
integrate only java14 switch expression, but ready to add the 21 null and 25 pattern. 
Import most code from the other switchexpression branch, but does not change java version.

The class to add the null  and pattern are still present, but the methods to add them are commented.

this allows to generate
https://github.com/glelouet/jcodemodel/blob/switchexpression17/jcodemodeltests/src/generated/javatest/com/helger/jcodemodel/tests/switchexpression/ESwitch.java
https://github.com/glelouet/jcodemodel/blob/switchexpression17/jcodemodeltests/src/generated/javatest/com/helger/jcodemodel/tests/switchexpression/BasicSwitch.java

from
https://github.com/glelouet/jcodemodel/blob/switchexpression17/jcodemodeltests/src/main/java/com/helger/jcodemodel/tests/switchexpression/SwitchExpressionTestGen.java

### DIFF
--- a/jcodemodel/src/main/java/com/helger/jcodemodel/JCodeModel.java
+++ b/jcodemodel/src/main/java/com/helger/jcodemodel/JCodeModel.java
@@ -732,6 +732,13 @@ public class JCodeModel implements Serializable
     }
     return aRefClass;
   }
+  
+  ///
+  /// reference a existing enum value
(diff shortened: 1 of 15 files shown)


glelouet:
yeah \o/

PR #145, opened 2026-05-24
text blocks

glelouet:
Text blocks are multiple text lines used as string.

This PR adds a JTextBlock class, that allows to append lines in a text block , with the following rules

1. newlines are used to split in multiple textblock lines, adding `a\nb` is the same  as adding `a` then `b`
2. triple parenthesis is escape
3. line-ending whitespace or tab is converted to octal whitespace to force keep the trailing whitespaces

What's more, the full text block can be added indentation, with selectable tab/space.
That should work even with single-line or any non-empty terminal line, eg `a\nb` with indent 2 **should** be exported as
```java
"""
  a
  b""".indent(2)
```

### DIFF
--- a/jcodemodel/src/main/java/com/helger/jcodemodel/JTextBlock.java
+++ b/jcodemodel/src/main/java/com/helger/jcodemodel/JTextBlock.java
@@ -34,9 +34,12 @@ public class JTextBlock implements IJExpression, Iterable<String> {
     return
         line
             // escape triple parenthesis
-            .replace("\"\"\"", "\\\"\"\"")
-            // trailing spaces/tabs are stripped : replace last with octal space
-            .replaceAll("[ \t]$", "\\\\040");
+            .replace("\"\"\"", "\\\"\"\\\"")
+            // trailing spaces/tabs are stripped : replace last space octal space (ascii d32
+            // = o040 )
+            .replaceAll(" $", "\\\\040")
+            // and last tab with octal tab (ascii d9 = o011 )
+            .replaceAll("\t$", "\\\\011");
   }
 
   private int indentSize = 0;
@@ -130,7 +133,7 @@ public class JTextBlock implements IJExpression, Iterable<String> {
       lastEmpty = line.isEmpty();
     }
     f.print(LIMITER);
-    if (!lastEmpty) {
+    if (!lastEmpty && indentSize > 0) {
       // if the last line is not empty, then the delimiter is not enough to enforce
       // the indent. So we add a call after that.
       f
--- a/jcodemodel/src/main/java/com/helger/jcodemodel/JTextBlock.java
+++ b/jcodemodel/src/main/java/com/helger/jcodemodel/JTextBlock.java
@@ -1,0 +1,143 @@
+package com.helger.jcodemodel;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.stream.Stream;
+
+import org.jspecify.annotations.NonNull;
+
+/// Represents a text block declaration, one line at a time.
+///
+/// [https://docs.oracle.com/en/java/javase/26/language/text-blocks.html]
+///
+@SuppressWarnings("serial")
+public class JTextBlock implements IJExpression, Iterable<String> {
+
+  public JTextBlock() {
+  }
+
+  public JTextBlock(String value) {
+    add(value);
+  }
+
+  /// convert a user line into several individual text block lines.
+  static Stream<String> formatLines(String line) {
+    return line.lines()
+        .map(JTextBlock::formatLine);
+
+  }
+
+  /// format a standard String line to make it a text block line.
+  static String formatLine(String line) {
+    return
+        line
+            // escape triple parenthesis
+            .replace("\"\"\"", "\\\"\"\"")
+            // trailing spaces/tabs are stripped : replace last with octal space
+            .replaceAll("[ \t]$", "\\\\040");
+  }
+
+  private int indentSize = 0;
+
+  public JTextBlock indentSize(int val) {
+    indentSize = val;
+    return this;
+  }
+
+  public int indentSize() {
+    return indentSize;
+  }
+
+  private char indentChar = ' ';
+
+  public JTextBlock indentChar(char val) {
+    if (val != ' ' && val != '\t') {
+      throw new UnsupportedOperationException("escape char must be space or tab");
+    }
+    indentChar = val;
+    return this;
+  }
+
+  public JTextBlock indentSpace() {
+    return indentChar(' ');
+  }
+
+  public JTextBlock indentTab() {
+    return indentChar('\t');
+  }
+
+  public char indentChar() {
+    return indentChar;
+  }
+
+  private List<String> lines = new ArrayList<>();
+
+  ///
+  /// transforms a line to make it fit to the text block syntax.
+  ///
+  /// @param line if null nothing happens
+  public JTextBlock add(String line) {
+    if (line != null) {
+      formatLines(line).forEach(lines::add);
+    }
+    return this;
+  }
+
+  /// shortcut to add a new line
+  public JTextBlock newline() {
+    lines.add("");
+    return this;
+  }
+
+  /// shortcut to add new lines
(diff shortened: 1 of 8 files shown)


glelouet:
made some tests, found bugs and fixed them.

Two issues remain : 

1. 

> 
> [ERROR] Forbidden method invocation: java.lang.String#replace(java.lang.CharSequence,java.lang.CharSequence) [Use StringHelper methods instead]
> [ERROR]   in com.helger.jcodemodel.JTextBlock (JTextBlock.java:37)

Could not find replacement for String.replace in StringHelpers. Fix @phax  please.

2. 
since I call formater.print, it adds a whitespace at the end. This is syntaxically same as not adding it, as trailing whitespaces are discarded, not a big issue but one notheless

produces
```java
    public static final String TWO_LINES_ENDTAB = """
    a	\011
    b	\011 """;
```

expected

```java
    public static final String TWO_LINES_ENDTAB = """
    a	\011
    b	\011""";
```

jcodemodel/src/main/java/com/helger/jcodemodel/JTextBlock.java
```diff
@@ -0,0 +1,146 @@
...
+        .map(JTextBlock::formatLine);
+
+  }
+
+  /// format a standard String line to make it a text block line.
+  static String formatLine(String line) {
+    return
+        line
+            // escape triple parenthesis
+            .replace("\"\"\"", "\\\"\"\\\"")
```
phax: Replacing one double quote is enough, so I think
```java
.replace("\"\"\"", "\\\"\"\"")
```
would be fine

jcodemodel/src/main/java/com/helger/jcodemodel/JTextBlock.java
```diff
@@ -0,0 +1,146 @@
...
+            // trailing spaces/tabs are stripped : replace last space octal space (ascii d32
+            // = o040 )
+            .replaceAll(" $", "\\\\040")
```
phax: I think this (and the tab replacement) could be made customizable - `boolean trimValue` or so?

jcodemodel/src/main/java/com/helger/jcodemodel/JTextBlock.java
```diff
@@ -0,0 +1,146 @@
...
+            // and last tab with octal tab (ascii d9 = o011 )
+            .replaceAll("\t$", "\\\\011");
+  }
+
+  private int indentSize = 0;
```
phax: Please move all vars to the top

jcodemodel/src/main/java/com/helger/jcodemodel/JTextBlock.java
```diff
@@ -0,0 +1,146 @@
...
+  public JTextBlock indentSize(int val) {
+    indentSize = val;
+    return this;
+  }
+
+  public int indentSize() {
+    return indentSize;
+  }
+
+  private char indentChar = ' ';
```
phax: Please move all vars to the top

jcodemodel/src/main/java/com/helger/jcodemodel/JTextBlock.java
```diff
@@ -0,0 +1,146 @@
...
+
+  public JTextBlock indentTab() {
+    return indentChar('\t');
+  }
+
+  public char indentChar() {
+    return indentChar;
+  }
+
+  private List<String> lines = new ArrayList<>();
```
phax: Please move all vars to the top

jcodemodel/src/main/java/com/helger/jcodemodel/JTextBlock.java
```diff
@@ -0,0 +1,146 @@
...
+
+  private static final String LIMITER = "\"\"\"";
+
+  @Override
+  public void generate(@NonNull IJFormatter f) {
+    f.print(LIMITER).newline();
+    String indent =
+        indentSize <= 0 || lines.isEmpty()
+            ? ""
+            : ("" + indentChar).repeat(indentSize);
```
phax: Please use `Character.toString (indentChar)` instead of `"" + indentChar`

jcodemodel/src/main/java/com/helger/jcodemodel/JTextBlock.java
```diff
@@ -0,0 +1,146 @@
...
+      }
+      f.print(indent).print(line);
+      first = false;
+      lastEmpty = line.isEmpty();
+    }
+    f.print(LIMITER);
+    if (!lastEmpty && indentSize > 0) {
+      // if the last line is not empty, then the delimiter is not enough to enforce
+      // the indent. So we add a call after that.
+      f
```
phax, line 212: This and the following 3 lines look odd

phax (changes requested):

jcodemodel/src/main/java/com/helger/jcodemodel/JTextBlock.java
glelouet: > Replacing one double quote is enough, so I think
> 
> ```java
> .replace("\"\"\"", "\\\"\"\"")
> ```
> 
> would be fine

No it's not. I did that at first then had tests not passing.

If your text is `"""` then the corresponding text block is
```java
"""
\""""""
```
and thus the 3 dquotes from index 1 to 3 are the end, with two remaining dquotes and a starting escaped quote.

I think I need to make more tests with ending "

edit : it's a matter of the last line ending with dquotes. I will replace only first an,d fix that for this specific case.

jcodemodel/src/main/java/com/helger/jcodemodel/JTextBlock.java
glelouet: if you want to. But the most important is, that by default they are not. A text block with `a   `  then `b\t' on the next line should by default have those values.

The issue with a param trimvalue is that it would require to untrim when set to false.

a middleground could be to have `addTrim` in addition to add, which does the trim

jcodemodel/src/main/java/com/helger/jcodemodel/JTextBlock.java
glelouet: That's the recommended way to do it, explained in the link.

When the last line is not empty, then the delimiter is added at the end. 
Since the *minimum* common indentation is stripped, this means
```java
"""
  a
  b"""
```
having 2 indentations on each line would be parsed as `a\nb` instead of expected `  a\n  b`

phax: I overlooked `String.indent` is an official method since Java 12 - this makes sense now

glelouet:
More bugs with double quotes fixed.
If quoting FIRST double quote, then `"""""` would be quoted as `\"""""` leading to the dquotes from 1 to 3 being delimiter.
If the LAST character of the last line  was an unescapoed doublequote, this would lead to quadruple quotes so error.  to avoid this, last char of last line is escaped if unescaped parentheses.

glelouet:
I think I'm a bit confused with the whole ident thing.

Since I escape the last space/tab, I may also need to ensure the first ones are not stripped.

this means, a preventStrip field.

Typically, I want to be able to write `  a  \n  b  ` and this means, either the result is the same (with preventStrip) , or the result is `a\nb` (without preventStrip). 

preventStrip would produce
```java
"""
\040 a \040
\040 b \040"""
```

while not preventstrip would produce
```java
"""
  a  
  b  """
```

glelouet:
ok I changed it.

Now default does not prevent removal of spaces by the parser. The *generated file* contains what you add, with the exception that the doublequotes are escapd when needed.

Calling `textblock.keepWhiteSpaces(true)` however will force the escape of the last character when whitespace ; and will force the escape of first char when whitespace and all line start with whitespace.

phax:
Thanks @glelouet  for your efforts - as usual :-)

glelouet:
I noticed at least one typo and maybe I don't need to escape all leading whitespaces (only first line), plus I think more doc would be useful regarding the indent part.

glelouet:
@phax How did you fix String.replace being rejected by forbiddenapis ?

phax:
I will deal with it - no worries. `StringReplace.method` instead

PR #147, opened 2026-05-27
minor+doc

glelouet:
Only first char is escaped when textblock requires to escape whitespaces, added small test


phax:
ready?

glelouet:
> ready?

please review and check mistakes :)

I did a pass so second commit but otherwise yes.

PR #151, opened 2026-05-29
Textblock fixes

glelouet:
Fixes  intermediate blank (typically empty) line still counting for escaping the first char (when keepwhitespaces)
Typically exporting `\n  a` was producing `\na`

Replaces the octal escape for space `\040` with `\s` for nicer visual


phax:
And `\011` can be replaced with `\t` which is also nicer and shorter. Will do it afterwards

glelouet:
> And `\011` can be replaced with `\t` which is also nicer and shorter. Will do it afterwards

You can edit/add commit to my files.

glelouet:
note : 
The bug was indeed present, as I added unit tests and they failed. So the unit tests were failing, but this  PR also has fix.

PR #155, opened 2026-06-05
Emods ijmodified

glelouet:
step forward in the effort to add modern-ish (come on, enumset are soo 2000 ) code. No removal, just adding simpler methods to do the same thing.

1. added the IJModified interface in modifiers/.  Allows to access the Emods with add, remove, list methods
2. sameish methods  implemented in JMods (except addEmods requires allowed set)
4. AbstractJClassContainer extends IJModified
5. JDefined class implements it by delegating to its jmods, checking for interface/class allowed emods
6. JDirectClass implements it by doing nothing
7. JVar implements IJModified
8. JFieldVar overrides the addemods to use allowed fields emods instead ; and to return JFieldVar instead of a JVar :P 
9. JMethod implements IJModified

The new addEmod method automatically discards invalid modifiers (like public for a local variable) and removes the ones excluded by those added : 

Typically calling
```java
jmethod.addEmods(EMod.PUBLIC,  EMod.SEALED, EMod.PROTECTED)
```
Will result in the jmethod having PROTECTED modifier, with PRIVATE and PUBLIC removed if they were present ; the addition of SEALED is discarded, other existing modifiers are kept.

side note : final and volatile should also be in the list of mutual exclusion. The code style wrecks if I add them so  I opt out (I added them, then realized there was too many changes so reverted the project and did it again. Took me 1h to try to fix it, before I gave up).
This means a variable should be removed volatile when set to final (which has no effect since vars can't be volatile ) ; and same for a field (which allows both, so effect)


glelouet:
instead of addEmods() it's now emod() with at least one (null allowed) param

PR #158, opened 2026-06-06
TextBlock noindent

glelouet:
fix the formatter adding arbitrary spaces/indentation in a text block

1. the base idea is to build on a StringBuilder when generating a textblock against a formatter, only appending the sb at the end, rather than directly append all lines.
2. This requires to have visibility of the newline property used. So IJFormatter, JFormatter, SourcePrintWriter all have a getter for that internal value (which in the case of JFormater calls the internal sourceprintwriter)
3. ran sh/install and a lot of things changed, no idea why. At least the output of the textblock is correct.

### DIFF
--- a/jcodemodel/src/main/java/com/helger/jcodemodel/IJFormatter.java
+++ b/jcodemodel/src/main/java/com/helger/jcodemodel/IJFormatter.java
@@ -89,6 +89,11 @@ public interface IJFormatter extends Closeable
    */
   @NonNull
   IJFormatter newline ();
+  
+  /**
+   * @return the internal String used for new lines.
+   */
+  String getNewLine();
 
   /**
    * Print a char into the stream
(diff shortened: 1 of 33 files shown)


glelouet:
ok, all those random space changes are because when I checkout the GenerateTestFiles, git replaces its starting spaces with tabs, including in the textblock ; but once placed in comment, the now-tab spacing is no more replaced with spaces because it's not starting anymore.

Its not a problem for the decoration spaces, which are discarded at compile time, but for that one line that is indented it becomes a issue.

glelouet:
okay I removed the smudge part from the git config script, this leads to less code changed.

`git ls-files -z | xargs -0 rm ; git checkout -- .` forced a reload of local files with new filter.

`git checkout HEAD -- **` should work too ? dunno, I ran several commands until `sh/cleaninstall && git status` showed me changes :P

phax:
Okay, sounds nice for me. Are you happy? Can I merge?

glelouet:
Yeah, I have a lot less changes to undo when committing so I'm pretty happy.
Yeah you can merge.

PR #160, opened 2026-06-08
Lazy init pattern

glelouet:
Make JFieldVar and JMethod constructor visible to construct them in the export of the lazy without adding them to the class

Add extraDeclarations to the JDefinedClass to add the lazy(s). JLazy auto adds itself in the jdc.

Add tests on the JLazy static method, and generation usage + generated tests in jcodemodeltests.


glelouet:
Note : the present implementation of the declare() may be changed later to use new java features (with literally lazyvalues )

PR #161, opened 2026-06-10
added type inference("var")

glelouet:
JVar relaxed non-null constraint on the type ; however fails if declaring a variable without type AND without init.

JFieldVar got this constraint instead, since class variables can't be var (only local method variables)

Same, JMethod.param() has a check on null since method params can't be var.


phax:
Looks good. Let me know when ready

glelouet:
I think it's ready, but I did not add tests because I don't see how to test this. I just added a generation that creates int and var, for me that's enough as the result is that it compiles (it did not at first).

I guess I could add `return variable;` but I just can't think of a correct test . So no test :P I guess they will come with bug tickets :D

Issue #156, opened 2026-06-06
Formatter settings


glelouet:
The JCMWriter should be able to parse several format of java code specification and apply them, so that when writting a JCM the output respects the specification.
For example, tabs or spaces ? Can we used hearts instead of stars in our javadoc newlines ? Add a newline before an opening curling bracket , or maybe ten ?
Do we wrap instructions in one line, same for expressions ?

Personally I wrap them when they go over 20chars and align the following ones on an indentation twice past the initial instruction, like
```java
// starting 2 space indent 
..var nonEmpty=list.stream()
// next part of the instruction is wrapped with 2 additional indentations (so 6 spaces) because the WHOLE instruction is >20 chars
.......filter(o->o!=null
// here there are two expressions, so the second one is again 2 additional indentations so 10 spaces
..........&& !o.isEmpty())
.......toList();
```

Those *choices* (alongside a lot of other ones) should be available.

glelouet:
since last PR, wrapping is now done, formatter is now avail, the rest need to be discussed (should we use ascii hearts or utf ones ?)

phax:
Depending on the selected character set ;-)

glelouet:
@phax so what's the plan now ?
Do you intend to merge the other PR this year or the next ?

Honestly your delay in taking into account my changes has been a huge letdown. To the point I was feeling like the project was abandoned. So I joke about it, but that's because I cant do anything else about it.

If you could check the eclipse formatter xml to be sure it satisfies your preferences, that could help. Because I'm fed up of having to check in each commit what I changed, and revert what was auto formatting implied
Load the project in eclipse, right click on jcm, settings, search "formatter", project-specific settings, import jcodemodel/meta/eclipse/formatter.xml, edit the "jcodemodel" profile that you just loaded.

phax:
@glelouet well, fair point - I can feel your pain. Obviously, jcodemodel is not near any of my priorities. You have way more interest in making jcodemodel "complete" and let it create beautiful code :-) Do you see a different way forward, so that your expectations are better fulfilled?

Lets ignore the formatting - I will take care of it...

glelouet:
> Do you see a different way forward, so that your expectations are better fulfilled?

Yes : lower my expectations :D 
But seriously, having formatting automated would help spending less time on such a tedious dimension :

> Lets ignore the formatting - I will take care of it...

I'm sorry Dave, I'm afraid I can't do that.
[...]

PR #163, opened 2026-06-25
Formatter settings, and multiple additions

glelouet:
# issues addressed

#156 
#164 
#165 
#166 

# Features added

## Local Variable list

The hierarchy of the JVar was changed, so creating a block variable now returns  a JBlockVar that allows to add additional vars at the same time.

```java
JBlockVar i = myBlock.declare(0, jcm.INT, "i");
JSameVar j = i.andVar("j");
```

generates

```java
int i, j;
```

this is also possible in a for loop, and in an instance field.

## For Loop init fix

a "for' loop in java mut be initialized by one of : 
 - nothing
 - a local variable alongside its children variables
 - a list of statements

The current implementation allows to mix both. I changed that to have only one of, throwing exception if both are used.

Tests show how this works.

## Catch multiple types

Once the vars hierarchy was changed, new classes could be added, including a JCatchVar for a catch clause, that allows to add additional types to the var. 

oracle doc for actual lub type of the catch clause : 
https://docs.oracle.com/javase/specs/jls/se25/html/jls-14.html#jls-14.20-510

The structure of a catch clause has been changed : 
Before, it contained a variable that could be null, the type of the exception ; calling the variable with a name could create it or throw an exception ; and during export the variable was created if missing. Now it only contains a variable, never null, and allows to change its name at will, or add new types to the variable list. This avoids printing a model having side effects, also it avoids throwing exception for no reason.

I removed the check for a JVar::setType to not receive null value, which was overdue anyhow.

Once the var of a catch block receives more than one type, its own returned type is null, for inspection purpose, since it can't link back to a JCM to reference Exception.

Tests show how the settings impact the multicatch formatting.

## Array init

Allows to generate

```java
int[] i= {0} ;
```

as specified in https://docs.oracle.com/javase/specs/jls/se25/html/jls-8.html#jls-VariableInitializer

For this, a new interface `IVariableInitializer` super of `IJExpression` is added, with single direct class `JArrayInit` .

This interface is used for variable declarations. A `JExpr.arrayInit` helper method directly calls the constructor.

In the concrete class, any *null* element is replaced by the `JExpr._null()` for ease . This way you can 

```java
myBlock.decl(STRINGCLASS.array(), "s", Jexpr.arrayInit(null, null);
```

which produces 

```java
String[] s = {null, null };
```

JExpr also has helper function for char, int, double.

Both the result and the format are tested.

## Formatter settings

This first step is to set settings at Jformatter level.

The class  `com.helger.jcodemodel.writer.FormatterSetings` will hold data for formatter settings. 
It has a `FormatterSettings configure(Consumer<FormatterSettings>)` method that returns this after accepting this. This method allows to chain several configuration, eg 
```java
var fs = new FormatterSettings()
  .configure(c->c.indent.useSpaces(3))
  .configure(c->c.wrap.disabled=true)
;
```

All the child settings fields (like indent, or wrap) are final public to allow fast access/change.

Then the basic rules for the settings are
1. keep same default behavior when possible. Default indent is 4 spaces, so use 4 spaces by default
2. add setter methods and those for often-performed. Typically a boolean field with default false will have a method to set it to true : wrap.disabled => wrap.setDisabled(value) + wrap.disabled()
3. most changing methods return the element being changed, for chaining.

The JCMWriter received a new field of this class. When creating a JFormatter, it transmits the settinbgs instead of only the indent String. The JFormatter field is final , but not the jcmwriter's.

## General Indentation options

settings.indent has 2 fields : 
 - `String string` is the indent String. Can be set using useSpaces(n) which defaults to 4 spaces, or useTabs(n) which defaults to 1 tab. The default value is `JCMWriter.DEFAULT_INDENT_STRING` to not change a thing.
 - `int tabSize` tells us how many chars is a tab column at most. This is useful to find the size of a line that contains tabs

### Wrapping settings

Those allow to choose how to wrap several parts of code declaration.

Several classes are shared among those options, this allows to define the wrapping among less classes. A `genericPrints` method was added in the JFormatter, that takes a collection of Objects as well as the separator and list-wrapping options.

The existing strategy is to wrap all params after first if >3 ; not wrap otherwise. 
Common wrap strategies consider several things, including the need for wrap (when the line gets over a specific size), the wrapping of first element, the indentation of the wrapped lines (existing is 1)

I try to keep it simple so I selected a few wrap methods : 
 - always wrap, including first element
 - never wrap, everything on the same  line
 - binary wrap : if the line is too big, wrap ALL elements, otherwise wrap none.
 - required : only wrap each specific element that would make the line too big
 - past3 : this is legacy method, wrap after first if more than 3 elements.

To detect the need for wrapping, I added a `wrap.lineCharacters = 80` option. 

Since there may be bugs, the full feature is locked behind a `wrap.disabled=false` option. When set to true, the generation of code *should* literally use legacy code.

Other than that, the list wrapping options can be specified the indentation on wrapping, as well as the requirement to wrap after or before the separator.

Then there was an issue : how to detect if the line is too big ? I had to use a "try something" approach

### Temporary context stack

I added a `addContextLayer` method to the JFormatter to add a new context on top of its internal stack. A context contains a StringBuilder, as well as parameters inherited from the previous one.
As  long as a context is present, writing to the formatter actually writes in the context. The context then has two methods `rollback() and commit()` , which set an internal `persist` flag then close it. Closing a context pops it off the context layers of the formatter, then write its content if internal persist flag set to true.
The `persist` flag can also be set manually, and the buffer implements autocloseable, so one can 
```java
try(var buffer = formatter.buffer().persistOnclose())
{
  /// add things to the context.
}
/// the buffer is auto closed when exiting the try, so persisted - unless manually rollbacked in the try block
```

Note that if a buffer is created on top of another one, the persist (or commit) will actually write in that underlying buffer. This is because, we can try things inside trying things ^^ 

Now with that done, the wrapping of a REQUIRED can simply try to add the next param, and if the line is too big it rollbacks, add a newline, then add the param again :) .

Same for BINARY, this tries to add all the params without newline, then if the line is too big it rollbacks, and add with ALWAYS instead.

Of course there is an added function to get the *current* line, which is present in the buffer and the JFormatter, just like lastChar is also added in the Buffer ; to make things simpler a `printDown` method is added in the JFormatter to select where to write, and a utlity method that counts the size of a line when using tabs with given column size .

### Formatting tests

Each formatting setting has its own package. for example `jcodemodel.test.format.method` for the method formatting settings.

In those packages,  several classes are generated with the same base generation, just different name and settings.. This means that when a change happens on a setting behavior, it's easy to track them. For example, I changed an internal method and forgot to add  a space after the comma, and it showed after a mvn install in the git status.

Those can't fail the build, they are here for tracking, as well as deciding if the result is fine in the initial dev phases. I found several bugs with them, so they may look useless in terms of tests but actually help.


glelouet:
I'm adding a small change to the JFormatter : the ability to buffer the writes in an intermediate StringBuilder.

This buffer can be commited, rollbacked, and if used in a try-with-resource is automatically discarded when exiting the block (unless commited before).

This allows to have optimist approach that goes for a simple thing, then commit if okay, rollback and redo it in another way otherwise.

glelouet:
I fixed a few bugs, added "tests" as those are more visual than actually checking things.

The result is the difference on the various produced classes (all with same methods, only class name and options change) https://github.com/phax/jcodemodel/pull/163/changes#diff-3c04635b2388974428723133096c9912cf755e4183ddd4d643cb11b478b17cc2

The test classes ( @TestJCM ) now injects a FormatterOptions params, to allow to manipulate it for tests.

@phax it's quite the mess so look at the result before you look at the code :D

glelouet:
Added ability to wrap the block of a method's body, starting with bracket. Allows to have bracket on new line with 

```java
		options.wrap.method.bracket
				.condition(EWrapWordStrategy.ALWAYS)
				.indent(0);
```

glelouet:
tried to implement the args part (so when invoking a method) but that would require too much copy paste. Basically needs to copypaste the whole code.
So I was pondering whether to add dedicated code in the JFormater.generate(List<JVar>). Turns out, jvars have a state() that is fixed for block variables , not correct for method params.
So I think we should have a dedicated JVarParam for them instead.  But then that's maybe too muc for that PR.
so I kept the params, but not effect. So far.

glelouet:
I think there was a bit of issues with vars.
Mainly, several time instead of formatter.var(jvar) they are hardcoded in very specific ways. 

So I changed it to have a new hierarchy for JVAR

 - JVar is the root part, with everything possible for a var. Later will be set to abstract. 
 - JFieldVar is kept as is for legacy reason
 - new package vars/ to contains the vars implementations :
 - JBLockVar extends JVar is the old JVar : a variable that is declared in a block. Literally no change
 - JArgVar for vars declared in the sig of a method. They can't be assigned value, can only be final, can't have null type.
 - JVarArgVar is the vararg last param. Sorry if the name is bad, feel free to change it. The difference is that the type is followed by "..." when  binding. So the previous class has a method bindType() called in bind() , overriden in JVarArgVar. 
 - JCatchVar is a var declared for a catch block. They can have multiple types, no init, can only be final, require a type.
 - JForEachVar is a var declared in a `for( int a : mycollection)` it *requires* an init, can only be final, do not require a type. Their binding (for var)  uses ":" instead of "=" for assignment.
 - The JCatch and JForEach classes therefore only use their variables' declare instead of rewriting the full block.

I did not touch jlambdaparam because I'm lazy.

An improvement is that the JCatchBlock now can have several types for a single catch.

glelouet:
I was tweaking things (literally making the var() at the formatter level, using a genetic method so the wrapping can be reused for other things like the catch multitype list) and realized there was no way to set several vars with same types and modifiers at once.

This is now possible, you can write fields, block, and for vars as

```java
// one of the three
JfieldVar myVar  = myClass.field(Jmod.public | Jmod.static, jcm.INT, "i");
JBlockVar myVar = mybBlock.var(JMod.final, jcm.INT, "i");
JBlockVar myVar = myJForLoop.init(JMod.NONE, jcm.INT, "i", JExpr.lit(0));

JSameVar mySameVar1 = myVar.AndVar("j"),
  mySameVar2 = myVar.andVar("k", JExpr.lit(50))
;
```

I did not test the "for" though, I assume it's bugged since the other also had bugs.
especially since the spec for a FOR is EITHER a var list with the main type at the first var ; OR a statement list ; and the present code allow to merge both ?
https://docs.oracle.com/javase/specs/jls/se8/html/jls-14.html#jls-14.14.1

glelouet:
Now the jforloop only allows to EITHER init a variable, which be later based upon to add sub variables, OR init expressions.

Again, the local variable part allows to add more dimensions to the base type, eg from an int[][] i, adding 2 dimensions can create int[][][][] j in the same init (see example)

What's more a dedicated option is present to wrap the init

source : 
https://github.com/glelouet/jcodemodel/blob/formatterOptions/jcodemodeltests/src/main/java/com/helger/jcodemodel/tests/format/forloop/ForLoopTestGen.java
results : 
https://github.com/glelouet/jcodemodel/blob/formatterOptions/jcodemodeltests/src/generated/javatest/com/helger/jcodemodel/tests/format/forloop/WrapInitRequired.java

several result because I change the formatter options between each.

glelouet:
@phax 
I think I'm done :)

Next step : AST parser to create JCM
Then plugin to load with AST parser .
Then more Formatter configuration.
Then we can use it to reformat all the code I submit with many changes which are just in format, like one-line if without brackets.

phax:
Alrighty, now this is a big one :-)

jcodemodel/src/main/java/com/helger/jcodemodel/expression/JArrayInit.java
```diff
@@ -0,0 +1,35 @@
...
+public class JArrayInit implements IVariableInitializer {
+
+  @NonNull
+  private final IVariableInitializer[] elements;
+
+  public JArrayInit(IVariableInitializer... elements) {
+    this.elements = convertElements(elements);
+  }
+
+  static IVariableInitializer[] convertElements(IVariableInitializer[] elements) {
```
phax: Any reason this is not `private`?

jcodemodel/src/main/java/com/helger/jcodemodel/expression/JArrayInit.java
```diff
@@ -0,0 +1,35 @@
...
+    if (elements == null) {
+      return new IVariableInitializer[] {};
+    }
+    return Stream.of(elements).map(e -> e != null ? e : JExpr._null()).toArray(IVariableInitializer[]::new);
+  }
+
+  @Override
+  public void generate(@NonNull IJFormatter f) {
+    f.print('{');
+    f.generable(List.of(elements), ",", f.settings().wrap.variables.array);
```
phax: If this is converted to a List when generating, we could directly store it as one???

jcodemodel/src/main/java/com/helger/jcodemodel/util/NullWriter.java
```diff
@@ -64,14 +63,14 @@ private NullWriter ()
   {}
 
   @Override
-  public void write (final char [] cbuf, final int off, final int len) throws IOException
```
phax, line 67: Please leave the `IOException` in - if somebody derives from `NullWriter`, it should be possible to throw this exception. Applies equally to the following methods

jcodemodel/src/main/java/com/helger/jcodemodel/vars/JForEachVar.java
```diff
@@ -0,0 +1,72 @@
...
+///  - type can be null. In that case, "var" is used.
+///  - name needs be provided. Can be relaxed later for "_"
+///  - The init expression **must** be null
+///  - a new **collection** property is used instead to store the iteration
+///
+public class JForEachVar extends JVar {
+
+  protected IJExpression collection;
+
+  public JForEachVar(boolean final_,
```
phax: I would prefer `isFinal` over `final_` for readability - thx

jcodemodel/src/main/java/com/helger/jcodemodel/vars/JSameVar.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+  }
+
+  @Override
+  public String separator() {
+    return ",";
+  }
+
+  static AbstractJType typeArray(AbstractJType type, int dim) {
+    while (dim > 0) {
```
phax, line 55: Please add a check `if (dum < 0) throw new IllegalArgumentException ("...");`

glelouet:
I was thinking the miniscle dabble in the other PR did not motivate you to start incorporating them, so I started making Mother Of All Pr (MOAP) to clean it up.

jcodemodel/src/main/java/com/helger/jcodemodel/writer/settings/Indent.java
```diff
@@ -0,0 +1,59 @@
...
+  public int tabSize = 4;
+
+  public Indent useSpaces(int nb) {
+    string = " ".repeat(nb);
+    return this;
+  }
+
+  /// defaults to 4 spaces
+  public Indent useSpaces() {
+    return useSpaces(4);
```
phax, line 24: This should use `tabSize` instead of `4`

jcodemodel/src/main/java/com/helger/jcodemodel/writer/settings/Wrap.java
```diff
@@ -0,0 +1,137 @@
...
+  /// many
+  /// indent.
+  ///
+  /// This is a generic idea for any one-time element, like the "implements" part
+  /// of a class, a method's return type, etc.
+  public static class WordWrapping {
+
+    public enum EWordWrapStrategy {
+      /// always wrap the element
+      ALWAYS(),
```
phax, line 20: Please remove the brackets if there are no parameters

jcodemodel/src/main/java/com/helger/jcodemodel/writer/settings/Wrap.java
```diff
@@ -0,0 +1,137 @@
...
+      NEVER(),
+      /// only wrap if otherwise would go over the line size.
+      REQUIRED();
+    }
+
+    /// when do we wrap this specific code generation
+    public EWordWrapStrategy condition = EWordWrapStrategy.NEVER;
+
+    /// when we wrap, how much do we indent the code
+    public int indent = 1;
```
phax, line 31: Should 1 be a "public static final" constant somewhere?

phax, line 31: Please never use public variables - always create getters and setters - thx

jcodemodel/src/main/java/com/helger/jcodemodel/writer/settings/Wrap.java
```diff
@@ -0,0 +1,137 @@
...
+
+    public WordWrapping condition(EWordWrapStrategy value) {
+      if (value != null) {
+        condition = value;
+      }
+      return this;
+    }
+
+    public WordWrapping indent(int value) {
+      indent = value;
```
phax, line 41: Add a check that value must be > 0

jcodemodel/src/main/java/com/helger/jcodemodel/writer/settings/Wrap.java
```diff
@@ -0,0 +1,137 @@
...
+      /// only the minimum number of elements.
+      REQUIRED(false),
+      /// once an element should be wrapped, all are.
+      BINARY(true),
+      /// wrap all if more than 3 elements ; first item never wrapped
+      PAST3(false);
+
+      public final boolean twoPasses;
+
+      EListWrapStrategy(boolean twoPasses) {
```
phax, line 69: Wouldn't it be smarter to provide an int parameter `numberOfPasses` for more verbosity?

jcodemodel/src/main/java/com/helger/jcodemodel/writer/settings/Wrap.java
```diff
@@ -0,0 +1,137 @@
...
+
+    public ListWrapping condition(EListWrapStrategy value) {
+      if (value != null) {
+        condition = value;
+      }
+      return this;
+    }
+
+    public ListWrapping indent(int value) {
+      indent = value;
```
phax, line 92: Value range check missing

jcodemodel/src/main/java/com/helger/jcodemodel/writer/settings/Wrap.java
```diff
@@ -0,0 +1,137 @@
...
+      return this;
+    }
+
+    public ListWrapping wrapAfterSep(boolean value) {
```
phax, line 96: Create an overload without a parameter, similar to `wrapBeforeSep()`

jcodemodel/src/main/java/com/helger/jcodemodel/writer/settings/Wrap.java
```diff
@@ -0,0 +1,137 @@
...
+
+    public ListWrapping wrapBeforeSep() {
+      return wrapAfterSep(false);
+    }
+
+  }
+
+  /// wrapping is required if an element would increase the line above
+  /// this number of characters
+  public int lineWidth = 80;
```
phax, line 114: Add a constant for defaults

jcodemodel/src/main/java/com/helger/jcodemodel/writer/FormatterSettings.java
```diff
@@ -0,0 +1,73 @@
...
+
+public class FormatterSettings {
+
+  //
+  // help configs
+  //
+
+  public static final Consumer<FormatterSettings> CONF_PHELGER = settings -> {
+    settings.indent
+        .useSpaces(4)
```
phax: it's 2 :-)

jcodemodel/src/main/java/com/helger/jcodemodel/writer/FormatterSettings.java
```diff
@@ -0,0 +1,73 @@
...
+        .tabSize(4);
```
phax: And 2. I think `.tabSize(4).useSpaces()` should do the trick based on the proposed fix

jcodemodel/src/main/java/com/helger/jcodemodel/writer/JCMWriter.java
```diff
@@ -154,17 +151,20 @@ public JCMWriter setNewLine (@NonNull @Nonempty final String sNewLine)
...
-  @NonNull
-  public JCMWriter setIndentString (@NonNull final String sIndentString)
+  /// set the options, only if params not null.
+  /// @param options if null, not set (no effect)
+  /// @return this
+  public JCMWriter withSettings (FormatterSettings settings)
   {
-    ValueEnforcer.notNull (sIndentString, "IndentString");
-    m_sIndentString = sIndentString;
+    if (settings != null)
```
phax, line 164: Suggest to keep the ValueEnforcer to make the API error visible

jcodemodel/src/main/java/com/helger/jcodemodel/writer/JCMWriter.java
```diff
@@ -89,10 +89,7 @@ public class JCMWriter
   /** The newline string to be used. Defaults to system default */
   private String m_sNewLine = DEFAULT_NEW_LINE;
 
-  /**
-   * String to be used for each indentation. Defaults to four spaces.
-   */
-  private String m_sIndentString = DEFAULT_INDENT_STRING;
+  private FormatterSettings m_oSettings = new FormatterSettings ();
```
phax: Please rename `m_o...` to `m_a...` - thx :-)

jcodemodel/src/main/java/com/helger/jcodemodel/writer/JFormatter.java
```diff
@@ -128,21 +123,17 @@ public boolean isAmbiguousIn (@NonNull final JDefinedClass aEnclosingClass)
 
       // special case where a generated type collides with a type in package
       // java.lang
-      if (aSingleRef._package () == JFormatter.this.m_aPckJavaLang)
+      if (aSingleRef._package () == m_aPckJavaLang)
       {
         // make sure that there's no other class with this name within the
         // same package
         for (final JDefinedClass aClass : aEnclosingClass._package ().classes ())
-        {
```
phax, line 136: here you are removing brackets - I don't understand your logic ;-)

jcodemodel/src/main/java/com/helger/jcodemodel/writer/JFormatter.java
```diff
@@ -253,7 +246,7 @@ static boolean isJavaLangClass (final String sSimpleClassName)
 
     final Boolean ret = RESERVERD_JAVA_LANG_NAME.get (sSimpleClassName);
     if (ret != null)
-      return ret.booleanValue ();
+      return ret;
```
phax, line 249: Please keep the manual unboxing - it's more verbose

jcodemodel/src/main/java/com/helger/jcodemodel/writer/JFormatter.java
```diff
@@ -264,7 +257,7 @@ static boolean isJavaLangClass (final String sSimpleClassName)
     {
       bIsJavaLang = false;
     }
-    RESERVERD_JAVA_LANG_NAME.put (sSimpleClassName, Boolean.valueOf (bIsJavaLang));
+    RESERVERD_JAVA_LANG_NAME.put (sSimpleClassName, bIsJavaLang);
```
phax, line 260: Same here - no auto-boxing pls

jcodemodel/src/main/java/com/helger/jcodemodel/writer/JFormatter.java
```diff
@@ -282,10 +275,8 @@ private AbstractJClass _getClassForImport (@Nullable final AbstractJClass aClass
     {
       AbstractJClass aRealClass = aClass;
       if (aRealClass instanceof JAnonymousClass)
```
phax, line 277: We can use the variable directly here

jcodemodel/src/main/java/com/helger/jcodemodel/writer/JFormatter.java
```diff
@@ -374,23 +373,84 @@ public List <AbstractJClass> getAllSorted ()
...
    * Writer associated with this {@link IJFormatter}
    */
   private final SourcePrintWriter m_aPW;
 
-  private char m_cLastChar = 0;
-  private boolean m_bAtBeginningOfLine = true;
+  /// contexts to write into instead of the printwriter.
+  ///
+  /// No context should be kept here after closed.
+  private final List <FormatterContext> contextLayers = new ArrayList <> ();
```
phax: PLease start with `m_a` - thx

jcodemodel/src/main/java/com/helger/jcodemodel/writer/JFormatter.java
```diff
@@ -374,23 +373,84 @@ public List <AbstractJClass> getAllSorted ()
...
+    }
+
+    @Override
+    public void append (String fullString, boolean resetLine, String appendLine, char lastChar)
+    {
+      m_aPW.print (fullString);
+      m_cLastChar = lastChar;
+      if (resetLine)
+      {
+        currentLine = new StringBuilder ();
```
phax, line 439: `.setLength (0)` might be more efficient

jcodemodel/src/main/java/com/helger/jcodemodel/writer/JFormatter.java
```diff
@@ -413,18 +473,20 @@ public List <AbstractJClass> getAllSorted ()
    *        Incremental indentation string, similar to tab value. May not be
    *        <code>null</code>.
    */
-  public JFormatter (@NonNull @WillCloseWhenClosed final SourcePrintWriter aPW, @NonNull final String sIndentString)
+  public JFormatter (@NonNull @WillCloseWhenClosed final SourcePrintWriter aPW,
+      @NonNull final FormatterSettings formatterSettings)
   {
     ValueEnforcer.notNull (aPW, "PrintWriter");
-    ValueEnforcer.notNull (sIndentString, "IndentString");
+    ValueEnforcer.notNull (formatterSettings, "formatterOptions");
```
phax: Change 2nd param from `formatterOptions` to `formatterSettings`

jcodemodel/src/main/java/com/helger/jcodemodel/writer/JFormatter.java
```diff
@@ -554,136 +624,144 @@ private static boolean _needSpace (final char c1, final char c2)
...
     }
     return this;
   }
 
+  @Override
   @NonNull
   public JFormatter print (@NonNull final String sStr)
   {
-    if (m_eMode == EMode.PRINTING && sStr.length () > 0)
+    if ((m_eMode == EMode.PRINTING) && (sStr.length () > 0))
```
phax, line 659: No nested brackets when only one boolean operation is used - thx

jcodemodel/src/main/java/com/helger/jcodemodel/JForEach.java
```diff
@@ -52,28 +54,18 @@
...
-    ValueEnforcer.notNull (aVarType, "VarType");
     ValueEnforcer.notNull (sVarName, "VarName");
     ValueEnforcer.notNull (aCollection, "Collection");
-
-    m_aMods = aMods;
-    m_aType = aVarType;
-    m_sVarName = sVarName;
-    m_aCollection = aCollection;
-    m_aLoopVar = new JVar (JMods.forVar (JMod.FINAL), m_aType, m_sVarName, aCollection);
+    m_aLoopVar = new JForEachVar((aMods.getValue() | JMod.FINAL) > 0, aVarType, sVarName, aCollection);
```
phax: The old code was easier to read. Can you please add something like `aMods.isFinal()` - thx

jcodemodel/src/main/java/com/helger/jcodemodel/JForLoop.java
```diff
@@ -48,116 +48,170 @@
...
 /**
  * For statement
  */
-public class JForLoop implements IJStatement
-{
-  private final List <IJObject> m_aInitExprs = new ArrayList <> ();
+public class JForLoop implements IJStatement {
+
+  // either a init var, or expressions
+  private JBlockVar initVar = null;
```
phax: No need to init with null

jcodemodel/src/main/java/com/helger/jcodemodel/JForLoop.java
```diff
@@ -48,116 +48,170 @@
...
+  }
+
+  public JBlockVar getInitVar() {
+    return initVar;
+  }
+
+  /// thow an exception if can't create a new var
+  protected void checkInitVar() {
+    if (initVar != null) {
+      throw new RuntimeException("a for loop can only have one type variable, this already has one");
```
phax: Never throw RuntimeException - use IllegalStateException instead

jcodemodel/src/main/java/com/helger/jcodemodel/JForLoop.java
```diff
@@ -48,116 +48,170 @@
...
+    }
+    if (!m_aInitExprs.isEmpty()) {
+      throw new RuntimeException(
```
phax: Never throw RuntimeException - use IllegalStateException instead

jcodemodel/src/main/java/com/helger/jcodemodel/JForLoop.java
```diff
@@ -48,116 +48,170 @@
...
+          "a for loop must have either variable declaration or expressions, this already has expressions");
+    }
+  }
+
+  /// @return true if we can init using expressions
+  protected void checkInitExpr() {
+    if (initVar != null) {
+      throw new RuntimeException(
```
phax: Never throw RuntimeException - use IllegalStateException instead

jcodemodel/src/main/java/com/helger/jcodemodel/JMethod.java
```diff
@@ -234,9 +238,9 @@ public JVar paramAtIndex (@Nonnegative final int nIndex) throws IndexOutOfBounds
    * @return New parameter variable of type {@link JVar}
    */
   @NonNull
-  public JVar param (final int nMods, @NonNull final AbstractJType aType, @NonNull final String sName)
+  public JArgVar param(final int nMods, @NonNull final AbstractJType aType, @NonNull final String sName)
   {
-    final JVar aVar = new JVar (JMods.forVar (nMods), ValueEnforcer.notNull(aType, "type"), sName, null);
+    final JArgVar aVar = new JArgVar((nMods | JMod.FINAL) > 0, ValueEnforcer.notNull(aType, "type"), sName);
```
phax, line 243: Same `isFinal()`

glelouet:
also I think I should do more unit testing, but I'm not sure how to test some so nope. Already did automatic code generation in tests to detect changes int he commit, so a review *should* catch out any change.

phax (changes requested):
I found a few things, but the PR is so huge, that I surely missed something ;-)

jcodemodel/src/main/java/com/helger/jcodemodel/expression/JArrayInit.java
glelouet: yes, to have it testable in unit tests. If not tested then I forgot , or changed the code several time until it did not need tests but may still need later.

jcodemodel/src/main/java/com/helger/jcodemodel/writer/JCMWriter.java
glelouet: it's not an error, it discards null so if you have automated code that creates (or load) config, but failed (for example because config is deactivated) then nothing happens.

for example when a plugin loads a config to apply it ;)

jcodemodel/src/main/java/com/helger/jcodemodel/writer/FormatterSettings.java
glelouet: crap then I also need to fix the auto format xml because I also set it back to 4 :P

jcodemodel/src/main/java/com/helger/jcodemodel/writer/settings/Wrap.java
glelouet: What for ? I don't think there will be an issue if someone uses -1 or Integer.MAX_VALUE.

jcodemodel/src/main/java/com/helger/jcodemodel/expression/JArrayInit.java
glelouet: There was more specific things before, like converting Double or Integer using Jexpr.lit, but I removed it. 
(Just because the result is one line, does not mean the code I wrote and removed was not 50 lines  :D )

jcodemodel/src/main/java/com/helger/jcodemodel/expression/JArrayInit.java
glelouet: maybe. `convertElements` could return a list instead. Not sure if worth.
Done anyhow.

jcodemodel/src/main/java/com/helger/jcodemodel/util/NullWriter.java
glelouet: no, null writer should not throw exception, because it does nothing. 
Something that derives nullwriter should not either.

I removed them because they sound like a bad practice. But ok I put them back

jcodemodel/src/main/java/com/helger/jcodemodel/vars/JSameVar.java
glelouet: hu, no ? If < 0 nothing happens, return type.

jcodemodel/src/main/java/com/helger/jcodemodel/writer/settings/Wrap.java
phax: But it feels wrong and is neither document nor separately handled

jcodemodel/src/main/java/com/helger/jcodemodel/writer/settings/Indent.java
glelouet: They are not related.

jcodemodel/src/main/java/com/helger/jcodemodel/writer/settings/Wrap.java
glelouet: parenthesis ? Well there was parameters but then I removed them.

glelouet: You can do that yourself ? I mean, can you modify the sources ?

jcodemodel/src/main/java/com/helger/jcodemodel/writer/settings/Wrap.java
glelouet: This class is just structure, no behavior.

jcodemodel/src/main/java/com/helger/jcodemodel/writer/settings/Wrap.java
glelouet: no, it's only used in the JFormatter.genericPrint for a specific reason (decoupling the try and direct execution). I added it because I had some bugs and it threw exception where I had made the wrong call, then it told me when I used other code before I use the genericPrint and forgot to split the executions.

I had code duplicated like at three places and therefor had bugs before I merged them in a single spot.

jcodemodel/src/main/java/com/helger/jcodemodel/writer/settings/Wrap.java
glelouet: why ?

jcodemodel/src/main/java/com/helger/jcodemodel/writer/settings/Wrap.java
glelouet: why should it ?

Those classes are just configuration. The default configuration is not supposed to be used, that's why I added things like glelouet() . Adding public static values also means not changing them unless on a major upgrade, because people referencing them should consider if the change won't break a thing.

jcodemodel/src/main/java/com/helger/jcodemodel/vars/JForEachVar.java
glelouet: done

jcodemodel/src/main/java/com/helger/jcodemodel/writer/settings/Wrap.java
glelouet: I don't get why, but okay.

jcodemodel/src/main/java/com/helger/jcodemodel/writer/settings/Wrap.java
glelouet: no.

jcodemodel/src/main/java/com/helger/jcodemodel/writer/FormatterSettings.java
glelouet: done

jcodemodel/src/main/java/com/helger/jcodemodel/writer/FormatterSettings.java
glelouet: 2 yes, but "fix" is wrong : tabspace is about existing indentation, typically a string containing tabs, not adding indentation.

jcodemodel/src/main/java/com/helger/jcodemodel/writer/JCMWriter.java
glelouet: yes, mo_ma

jcodemodel/src/main/java/com/helger/jcodemodel/JForEach.java
glelouet: you mean replace with ? Yeah, makes sense.

jcodemodel/src/main/java/com/helger/jcodemodel/JForLoop.java
glelouet: is it a problem though ?

done and renamed to m_aInitVar

jcodemodel/src/main/java/com/helger/jcodemodel/JMethod.java
glelouet: nope, it's an int here.

jcodemodel/src/main/java/com/helger/jcodemodel/writer/settings/Wrap.java
glelouet: yes, because it's exactly that : not documented, not separately handled. Those are just structure, what is done with them depends on the implementation. Maybe later negative value means from the max size ?

jcodemodel/src/main/java/com/helger/jcodemodel/writer/JFormatter.java
glelouet: I tried to make several passes on the eclipe formatter, to stick to what was existing.
But I did not reformat the whole project each time - so maybe there are things that were changed with a previous setting I tried, and then I did not format back with a newer one.

My *personal* settings is to **always** have brackets around blocks. I lost days of debuging to adding code after an if without realizing it was not in the block. Several times.

For me, NOT adding brackets in a block is a bug source.
But I get that *you* may have a different opinion, so I changed the eclipse auto formatter that I export in the project.
At first I used with "only required" but then a if(X) switch(Y) that was 10 lines long and without bracket, so I changed to "unless single line return or throws"

jcodemodel/src/main/java/com/helger/jcodemodel/writer/JFormatter.java
glelouet: why no auto boxing ?

jcodemodel/src/main/java/com/helger/jcodemodel/writer/JFormatter.java
glelouet: that's not my code though ?
if you have ideas for improvement, please make a separate PR.

jcodemodel/src/main/java/com/helger/jcodemodel/writer/JFormatter.java
glelouet: done

jcodemodel/src/main/java/com/helger/jcodemodel/writer/JFormatter.java
glelouet: maybe ? I let you do that later .

jcodemodel/src/main/java/com/helger/jcodemodel/writer/JFormatter.java
glelouet: yeah, that lets you have a glance on how things changed

jcodemodel/src/main/java/com/helger/jcodemodel/writer/JFormatter.java
glelouet: parentheses ? Sure, revert it. I don't even know why it was changed.

glelouet:
I :+1:  the requests I did in commit, :-1: those that I don't want to have incorporated, the rest I don't know.

jcodemodel/src/main/java/com/helger/jcodemodel/writer/settings/Wrap.java
phax: haha - then I will do it

PR #176, opened 2026-07-23
[4.3.0] [m] Deprecate root jfieldvar

### LINKED ISSUE #173: /JFieldVar marked as deprecated
merged on next MINOR release

glelouet:
change minor (require new minor release)

depends (embarks) #175 
closes #173


jcodemodel/src/main/java/com/helger/jcodemodel/JFieldVar.java
```diff
@@ -90,12 +93,14 @@ public JFieldVar (@NonNull final JDefinedClass aOwnerClass,
   /**
    * @return The owning class. Never <code>null</code>.
    */
+  @Deprecated(forRemoval = true, since = "4.3.0")
```
phax: Well, the method names should not be deprecated - effectively its only the constructor and the class.

phax (changes requested):

jcodemodel/src/main/java/com/helger/jcodemodel/JFieldVar.java
glelouet: lol, I deprecated the wrong one :P

glelouet: actually it's save actions that added deprecated on all public methods. crahp.

phax: I assumed this. If possible try to turn off that specific save action

glelouet: I did it, as you can see the cleanup was changed to have it removed, 

https://github.com/phax/jcodemodel/pull/176/changes/120a2a7966fd059bb35bde762b3764845649c3c6#diff-9c5202a180f0e1f846ac307a432e16f75b5fcad52a3f396c5bd0d2e6d1b1112eR39

meanwhile I deactivated my "additional actions" because they were conflicting (replacing) the cleanup actions, without a way to have them exported.

phax: Then you can disable the save action "wrap in paranthesis" as well ;-)

phax: Ah you disabled them all - okay

glelouet: I disabled all "additional actions" so that the actions are not in duplicate between "clean up" profile and "save actions".
Yeah it's a complete mess, I agree with myself on that.

glelouet:
@phax this one needs to be integrated too.

I changed the version to 4.4.0 in case you bump the minor in the next release. If not, revert the last commit after merging.

glelouet:
@phax  nooo you missed this one :p

phax:
Okay - we will make a 4.4.1

PR #179, opened 2026-07-24
using SourceVersion to detect java keywords and valid identifiers

glelouet:
Did not remove the deprecated code. Should we ?

### DIFF
--- a/jcodemodel/src/main/java/com/helger/jcodemodel/JJavaName.java
+++ b/jcodemodel/src/main/java/com/helger/jcodemodel/JJavaName.java
@@ -40,8 +40,7 @@
  */
 package com.helger.jcodemodel;
 
-import java.util.HashSet;
-import java.util.Set;
+import javax.lang.model.SourceVersion;
 
 import org.jspecify.annotations.NonNull;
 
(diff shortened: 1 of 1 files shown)


phax:
Class `SourceVersion` is a JDK/JRE thing - it's e.g. not available on Android. As I know some users are using jcodemodel to create Android code, I would vouch against this PR

glelouet:
caveats : 
 - does not allow to write for lower java version with less keywords. However based on specific java version will stick to that java version.
 - does not allow to write for higher version. For example if project runs on j17 but a new keyword is added in j19, then this will *can* write invalid j19 code. This was already the case before, so not a regression.

glelouet:
Does not work for android

glelouet:
at first I wanted to be sure the keywords were correct wrt the spec, then I found that I could list them from oracle's specs ; even better, build them using JCM by release !!
But then I realized it was already exatly the case in SourceVersion, so use it instead of rewriting the wheel.
I guess the wheel it is, then :D

phax:
Nice idea indeed - I wasn't aware of the class myself. Are there any keywords we can takeover or copy?

glelouet:
I did not check all (because oracle list is vertical first) but could not find a difference. 

I found that the existing code (mainly, the old-fashioned Set with static{} init, that could easily be replaced by =Set.of() nowadays) was present in other places, and the issue I have is that it does not have feature-specific  list  ; though not a problem now since last keyword is _ in j9 so … not really impacting.

Yes, I am still at the stage of identifiers in the spec :P

phax:
Okay, so you are running the spec through - nice :-) Good luck with that and thanks for the effort you put into it.

PR #181, opened 2026-07-24
added selective representation of int (bin, dec, hex, oct)

glelouet:
JAtomInt has new property representation to select output format.

Those are per-instance and not at writer level because some int may be written hex, other dec, in the same project or even class.

Added chaining assignment and tests

### DIFF
--- a/jcodemodel/src/main/java/com/helger/jcodemodel/JAtomInt.java
+++ b/jcodemodel/src/main/java/com/helger/jcodemodel/JAtomInt.java
@@ -42,6 +42,8 @@ package com.helger.jcodemodel;
 
 import static com.helger.jcodemodel.util.JCHashCodeGenerator.getHashCode;
 
+import java.util.function.IntFunction;
+
 import org.jspecify.annotations.NonNull;
 
 import com.helger.base.equals.EqualsHelper;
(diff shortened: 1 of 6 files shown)


glelouet:
@pĥax var names are wrong, because I was changing them. Need to settle for correct names.

bodyFormatter instead of the representer ?

glelouet:
convert to draft does not work. I forgot to add thousand separators.

phax (approved):

phax:
The String representation via `to...String` never contains a thousand separators - only when using NumberFormat

glelouet:
> The String representation via `to...String` never contains a thousand separators - only when using NumberFormat

yes.

I want to be able to produce -1__2__3 (for -123, with separator every 1 char, and separator size=2) 

I know this is an extremely retarded case, bus using sep for 1000 is usual so may as well do it correctly.

glelouet:
as tested, 

```java
new JAtomInt (-10).separatorSize (1).separateEvery (3).binary()
```
would produce 

> -0b1_010

(separator size is default 1, every is default 0)

@phax don't hesitate to request more thorough tests (so couple of vars to test) , because the code is bit complex so it may break from an oversight

phax:
Can you please create some example Java code with it, that adds all different variants together - to make sure the representation is suitable for effective Java code.

glelouet:
good point.

glelouet:
@phax done.

phax:
Super thanks - looks good. Unfortunately I released without it - will be part of 4.3.1 then

glelouet:
Again, I don't have an opinion on what should be in a release. I try to keep the dependencies clear so if this has none, then … well , whatever ?

BTW I added labels *minor* and *patch* to know when a PR/issue impends a semver minor/patch change.
Typically this adds feature so is minor, while the "no var for class name" is only fix so a patch.

phax:
Got that. Nevertheless want to keep you posted.

The labels are a good idea

PR #182, opened 2026-07-24
Integer representation specification

glelouet:
# Objective, constraints and explanations

A previous PR already embarked the "base" feature, but this one is cleaner/more detailed, especially regarding sharing settings.

This PR aims at providing ways to represent integer/long format in the generated code.

When we want to write an int, for example "404", with `JExpr.lit(404)` , we may also want to specify how this value will be written in the resulting code, eg
1. the base :  "404" in decimal, "0624" in octal, "0x194" in hexadecimal
2. add the positive leading sign , so "+404"
3. padd for non-decimal base as in "0x00000194"
4. in the case of hex should we upper/lower case the letters and the "x" in the prefix ?
5. Then should we add "_" separators, and where ?

In short, "+4_0_4" is just as valid as "0000624".
This set of choices is a "representation" of integers. It's not in the formatting settings, because formatting is project-wide, while those can change on a class or even field basis.

The main use case of those representations are defining them, and using them in int/long fields (a long is just an int with tailing l/L), therefore the building of one received special attention, with default representations available, both to use directly or to be based on.
As constraints, the default representation of an int should produce the same, when possible, as before ; and should impend no performance hit. 

Since creating user-specific representation should not modify the default values, they are all records to be read only. Updating a representation's value actually create a new representation with that value updated. The JAtomInt and JAtomLong can be set their representation, and provide a few methods to update specific fields though it replaces the internal representation with an updated copy.

Those representation start with a base, for example octal, then add options, in the order : force positive sign, padding, use lower caps letters, use separators.

# Features

## Format class

A new `IntegerRepresentation` record contains information to represent int and long in the code.

### Creation

A few default representation are available ; you are recommended to create your own that you use in your own library, since the default ones are not designed to be core part of the lib (they are arbitrary choices so not as stable as choice presentation).

```java
// creation

// use default one, no separator
IntegerRepresentation myIR = IntegerRepresentation.DEFAULT;

// or default decimal one eg 1_000_000L
myIR = IntegerRepresentation.DEC;

// default hexa so 0xAB_CDL
myIR = IntegerRepresentation.HEX;

//specify all params
myIR = new IntegerRepresentation (false,
                                                                                 false,
                                                                                 EIntegerBase.DECIMAL,
                                                                                 0,
                                                                                 null,
                                                                                 0,
                                                                                 1,
                                                                                 true);
```

### Usage

JAtomInteger and JAtomLong have that new field. They extend a common class that allows to manipulate those fields internally, replacing the representation each time.

```java
// usage

var i = new JAtomInt(42, myIR);
i = new JAtomInt(42).representation(myIR)
// change the base to octal
i.octal()
// and add a separator every 3 char
.separateEvery(3)
// force the addition of +
.positiveSign(true);
```

### Formatting process

1. formatting an integer/long uses a base and creates sign, prefix, body, suffix
7. sign only present if negative value or representation's `positiveSign` is true 
8. prefix depends on the base selected. The representation's `prefixUpper` indicates if use upper or lower case prefix
9. body is converted from the base , then padded , then separated
10. suffix is only for LONG .  The representation's `suffixUpper` indicates if use 'L' or 'l'
11. padded only applied for bases that use it. Decimal does not allow padding.
12. body separation is either performed with a separator format, or if null using the `separateEvery` and `separateSize` fields.
13. separator format indicates before which char, and how many, separators to add : each '_' indicates one separator to insert, starting from the end.
14. separateEvery indicates after how many body chars to add separator, and separateSize how many separator each time.

### Padding example

(those are the tests)

42 padded 5 is
 - 0b101010 (no padding needed here)
 - 42 (can't pad decimal)
 - 0x0002a (padding in hex)
 - 000052  (padding in octal)

###  Separator Format example

1000 decimal with format : 
 - "X_X" produces "100_0" : only adds a separator before the before-last charater
 - "_xx__x" produces "1_00__0"

### DIFF
--- a/jcodemodel/src/main/java/com/helger/jcodemodel/literals/AIntegerRepresented.java
+++ b/jcodemodel/src/main/java/com/helger/jcodemodel/literals/AIntegerRepresented.java
@@ -1,0 +1,98 @@
+package com.helger.jcodemodel.literals;
+
+/// Something that has an IntegerRepresentation to update.
+/// 
+/// Its abstract because it's just a tooling class to extend.
+/// 
+/// @param T must be declaring class, eg `class A extends AIntegerRepresented<A>`
+public abstract class AIntegerRepresented <T extends AIntegerRepresented <T>>
+{
+
+  protected IntegerRepresentation representation = IntegerRepresentation.DEFAULT;
+
+  @SuppressWarnings ("unchecked")
+  protected T self ()
+  {
+    return (T) this;
+  }
+
+  public IntegerRepresentation representation ()
+  {
+    return representation;
+  }
+
+  /// change the internal representation to the provided one
+  ///
+  /// @return this
+  /// @see [IntegerRepresentation] for the meaning of individual fields
+  public T representation (IntegerRepresentation representation)
+  {
+    if (representation != null)
+      this.representation = representation;
+    return self ();
+  }
+
+  /// change the internal representation to match the request
+  ///
+  /// @return this
+  /// @see [IntegerRepresentation] for the meaning of individual fields
+  public T binary ()
+  {
+    return representation (representation.base (IntegerBase.BINARY));
+  }
+
+  /// change the internal representation to match the request
+  ///
+  /// @return this
+  /// @see [IntegerRepresentation] for the meaning of individual fields
+  public T decimal ()
+  {
+    return representation (representation.base (IntegerBase.DECIMAL));
+  }
+
+  /// change the internal representation to match the request
+  ///
+  /// @return this
+  /// @see [IntegerRepresentation] for the meaning of individual fields
+  public T hex ()
+  {
+    return representation (representation.base (IntegerBase.HEX));
+  }
+
+  /// change the internal representation to match the request
+  ///
+  /// @return this
+  /// @see [IntegerRepresentation] for the meaning of individual fields
+  public T octal ()
+  {
+    return representation (representation.base (IntegerBase.OCTAL));
+  }
+
+  /// change the internal representation to match the request
+  ///
+  /// @return this
+  /// @see [IntegerRepresentation] for the meaning of individual fields
+  public T separatorSize (int size)
+  {
+    return representation (representation.separatorSize (size));
+  }
+
+  /// change the internal representation to match the request
+  ///
+  /// @return this
+  /// @see [IntegerRepresentation] for the meaning of individual fields
+  public T separateEvery (int every)
+  {
+    return representation (representation.separateEvery (every));
+  }
+
+  /// change the internal representation to match the request
+  ///
+  /// @return this
+  /// @see [IntegerRepresentation] for the meaning of individual fields
+  public T padding (int padding)
+  {
+    return representation (representation.padding (padding));
+  }
+
+}
--- a/jcodemodel/src/main/java/com/helger/jcodemodel/literals/AIntegerRepresented.java
+++ b/jcodemodel/src/main/java/com/helger/jcodemodel/literals/AIntegerRepresented.java
@@ -63,7 +63,7 @@ public abstract class AIntegerRepresented <T extends AIntegerRepresented <T>>
   ///
   /// @return this
   /// @see [IntegerRepresentation] for the meaning of individual fields
-  public T hex ()
+  public T hexadecimal ()
   {
     return representation (representation.base (EIntegerBase.HEXADECIMAL));
   }
--- a/jcodemodel/src/main/java/com/helger/jcodemodel/literals/AIntegerRepresented.java
+++ b/jcodemodel/src/main/java/com/helger/jcodemodel/literals/AIntegerRepresented.java
@@ -1,6 +1,8 @@
 package com.helger.jcodemodel.literals;
 
-/// Something that has an IntegerRepresentation to update.
+import org.jspecify.annotations.NonNull;
+
+/// Something that has an IntegerRepresentation to update, in practice only JAtomInt and JAtomLong
 /// 
 /// Its abstract because it's just a tooling class to extend.
 /// 
@@ -8,6 +10,7 @@ package com.helger.jcodemodel.literals;
 public abstract class AIntegerRepresented <T extends AIntegerRepresented <T>>
 {
 
+  @NonNull
   protected IntegerRepresentation representation = IntegerRepresentation.DEFAULT;
 
   @SuppressWarnings ("unchecked")
@@ -24,7 +27,7 @@ public abstract class AIntegerRepresented <T extends AIntegerRepresented <T>>
   /// change the internal representation to the provided one
   ///
   /// @return this
-  /// @see [IntegerRepresentation] for the meaning of individual fields
+  /// @param representation if null, nothing changes.
   public T representation (IntegerRepresentation representation)
   {
     if (representation != null)
@@ -32,73 +35,70 @@ public abstract class AIntegerRepresented <T extends AIntegerRepresented <T>>
     return self ();
   }
 
-  /// change the internal representation to match the request
+  /// change the internal representation to show positive sign
   ///
   /// @return this
-  /// @see [IntegerRepresentation] for the meaning of individual fields
   public T positiveSign (boolean positiveSign)
   {
     return representation (representation.positiveSign (positiveSign));
   }
 
-  /// change the internal representation to match the request
+  /// change the internal representation to use binary base
   ///
   /// @return this
-  /// @see [IntegerRepresentation] for the meaning of individual fields
   public T binary ()
   {
     return representation (representation.base (EIntegerBase.BINARY));
   }
 
-  /// change the internal representation to match the request
+  /// change the internal representation to use decimal base
   ///
   /// @return this
-  /// @see [IntegerRepresentation] for the meaning of individual fields
   public T decimal ()
   {
     return representation (representation.base (EIntegerBase.DECIMAL));
   }
 
-  /// change the internal representation to match the request
+  /// change the internal representation to use hexadecimal base
   ///
   /// @return this
-  /// @see [IntegerRepresentation] for the meaning of individual fields
   public T hexadecimal ()
   {
     return representation (representation.base (EIntegerBase.HEXADECIMAL));
   }
 
-  /// change the internal representation to match the request
+  /// change the internal representation to use octal base
   ///
   /// @return this
-  /// @see [IntegerRepresentation] for the meaning of individual fields
   public T octal ()
   {
     return representation (representation.base (EIntegerBase.OCTAL));
   }
 
-  /// change the internal representation to match the request
+  /// change the internal representation to use a fixed separator size (the number of character
+  /// BETWEEN
+  /// each separated group), used only when **NO** separator format is provided
   ///
   /// @return this
-  /// @see [IntegerRepresentation] for the meaning of individual fields
   public T separatorSize (int size)
   {
     return representation (representation.separatorSize (size));
   }
 
-  /// change the internal representation to match the request
+  /// change the internal representation to use a fixed separator distance (the maximum number of
+  /// character IN
+  /// a separated group), used only when **NO** separator format is provided
   ///
   /// @return this
-  /// @see [IntegerRepresentation] for the meaning of individual fields
   public T separateEvery (int every)
   {
     return representation (representation.separateEvery (every));
   }
 
-  /// change the internal representation to match the request
+  /// change the internal representation to use a padding value. The padding is not used for decimal
+  /// base, since leading "0" makes an octal.
   ///
   /// @return this
-  /// @see [IntegerRepresentation] for the meaning of individual fields
   public T padding (int padding)
   {
     return representation (representation.padding (padding));
--- a/jcodemodel/src/main/java/com/helger/jcodemodel/literals/AIntegerRepresented.java
+++ b/jcodemodel/src/main/java/com/helger/jcodemodel/literals/AIntegerRepresented.java
@@ -65,7 +65,7 @@ public abstract class AIntegerRepresented <T extends AIntegerRepresented <T>>
   /// @see [IntegerRepresentation] for the meaning of individual fields
   public T hex ()
   {
-    return representation (representation.base (EIntegerBase.HEX));
+    return representation (representation.base (EIntegerBase.HEXADECIMAL));
   }
 
   /// change the internal representation to match the request
--- a/jcodemodel/src/main/java/com/helger/jcodemodel/literals/AIntegerRepresented.java
+++ b/jcodemodel/src/main/java/com/helger/jcodemodel/literals/AIntegerRepresented.java
@@ -38,7 +38,7 @@ public abstract class AIntegerRepresented <T extends AIntegerRepresented <T>>
   /// @see [IntegerRepresentation] for the meaning of individual fields
   public T binary ()
   {
-    return representation (representation.base (IntegerBase.BINARY));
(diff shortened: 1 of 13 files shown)


phax:
I really don't think it is useful to provide all this customizability. Putting effort in creating code that is optimized for people but solely read by compilers is imho not a big priority...

glelouet:
It depends. If you work on custom int format, you want to have eg the mask for the different part avail.
Say the format is SRC8_DEST8_BDY8_CRC4_TTL4 then you want to have the body mask
BDY_MASK= 000_00_77_0_0 ;

Also if you want to parse data to build other code from it, it may be better to parse the data faithfully.

glelouet:
Anyhow, it's good it's not released, because I think it needs more thinking.

I will try to have common interface-ish for all number AtomX classes. Even though for example you have only dec and hex representation for fp.

I think I will go for a more complex internal field to represent the formatting, shared for long/int, and another shared for float/double . So IntegerLiteralFormat / FloatingPointLiteralFormat .

phax:
But try to keep the default case (no formatting, decimal output) as efficient as possible - thx

glelouet:
ok so here is what I want :

 - be able to share the representation among several JAtomInt 
 - and also among JAtomLong (as in, same class for both)
 - while still able to modify the JAtomInt individually

So for me it's best to have a record holding the representation params, including the enum base (hex, dec, bin, oct) , but have the already present methods visible on the JAtom to change the record to a new one.
This way we can directly set the whole params, or change then one by one ; and changing them won't impact the instances they are already assigned to.

glelouet:
Much better this way. We can update the representation internally without changing the signature.
I had an awful lot of copypasta between int and long so I just made a common class. Thinking of it, could be an interface with default values. edit : no, it can't.

phax:
fyi: I am on vacation next week - so no expectations towards me pls - thx :-)

glelouet:
> expectations

I did not expect you would review all the PRs so fast. gladly surprised.

Also it's nice you took the formatting. Will help me focus on the important things.

But yes, I don't expect anything, except that you take care of yourself during your vacations :)

glelouet:
I made a pass on the tests, ready to review.

jcodemodel/src/main/java/com/helger/jcodemodel/literals/AIntegerRepresented.java
```diff
@@ -0,0 +1,107 @@
...
+  @NonNull
+  protected IntegerRepresentation representation = IntegerRepresentation.DEFAULT;
+
+  @SuppressWarnings ("unchecked")
+  protected T self ()
+  {
+    return (T) this;
+  }
+
+  public IntegerRepresentation representation ()
```
phax, line 23: I would prefer the annotations on the method return types and method parameters. If not, then not

jcodemodel/src/main/java/com/helger/jcodemodel/literals/AIntegerRepresented.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  /// change the internal representation to the provided one
+  ///
+  /// @return this
+  /// @param representation if null, nothing changes.
+  public T representation (IntegerRepresentation representation)
+  {
+    if (representation != null)
+      this.representation = representation;
```
phax, line 35: `else this.representation = null`?

jcodemodel/src/main/java/com/helger/jcodemodel/literals/IntegerRepresentation.java
```diff
@@ -0,0 +1,216 @@
...
+
+  public IntegerRepresentation
+  {
+    Objects.requireNonNull (base);
+  }
+
+  // copier
+
+  /// Intermediate mutable class for mutate a record. Waiting for java withers …
+  public static class Copier
```
phax, line 58: Would we spare that, when we change the record to a regular class?

jcodemodel/src/main/java/com/helger/jcodemodel/literals/IntegerRepresentation.java
```diff
@@ -0,0 +1,216 @@
...
+    ir.separateEvery = 4;
+  });
+
+  //
+  // mutators
+  //
+
+  public IntegerRepresentation positiveSign (boolean positiveSign)
+  {
+    return with (ir -> { ir.positiveSign = positiveSign; });
```
phax: All these methods could benefit from `if (ir.positiveSign == positiveSign) return this;` - to avoid unnecessary copying

jcodemodel/src/main/java/com/helger/jcodemodel/literals/EIntegerBase.java
```diff
@@ -0,0 +1,228 @@
...
+    this.allowBodyLeadingSep = allowBodyLeadingSep;
+  }
+
+  /// @return sb
+  public StringBuilder represent (int i,
+                                  StringBuilder sb,
+                                  IntegerRepresentation f)
+  {
+    boolean neg = i < 0;
+    i = neg ? -i : i;
```
phax: Please assign a local variable - parameters should always be final

jcodemodel/src/main/java/com/helger/jcodemodel/literals/EIntegerBase.java
```diff
@@ -0,0 +1,228 @@
...
+    return sb;
+  }
+
+  /// @return sb
+  public StringBuilder represent (long l,
+                                  StringBuilder sb,
+                                  IntegerRepresentation f)
+  {
+    boolean neg = l < 0;
+    l = neg ? -l : l;
```
phax: Parameter should always be final

jcodemodel/src/main/java/com/helger/jcodemodel/literals/EIntegerBase.java
```diff
@@ -0,0 +1,228 @@
...
+    this.prefixUpperCased = prefix.toUpperCase (Locale.ROOT);
+    this.intFormat = intFormat;
+    this.longFormat = longFormat;
+    this.enableBodyUpper = enableBodyUpper;
+    this.enablePadding = enablePadding;
+    this.allowBodyLeadingSep = allowBodyLeadingSep;
+  }
+
+  /// @return sb
+  public StringBuilder represent (int i,
```
phax, line 64: To me, this method (and the long variant) would better suit in `IntegerRepresentation` - otherwise we have some weird circle dependency between them

jcodemodel/src/main/java/com/helger/jcodemodel/literals/EIntegerBase.java
```diff
@@ -0,0 +1,228 @@
...
+        for (int formatIndex = sepFormat.length () - 2, sourceIndex = source.length () - 2; formatIndex >= 0 ||
+          sourceIndex >= 0; formatIndex--)
+        {
+          if (formatIndex < 0)
+          {
+            for (int i = sourceIndex; i >= 0; i--)
+              reversed.append (source.charAt (i));
+            break;
+          }
+          else
```
phax, line 190: No need for else

jcodemodel/src/main/java/com/helger/jcodemodel/literals/EIntegerBase.java
```diff
@@ -0,0 +1,228 @@
...
+          {
+            if (sepFormat.charAt (formatIndex) == SEP_CHAR)
+              if (allowLeadingSep || sourceIndex >= 0)
+                reversed.append (SEP_CHAR);
+              else
+                break;
+            else
```
phax, line 197: But the block above the else in brackets - this could be very misleading otherwise ;-)

jcodemodel/src/test/java/com/helger/jcodemodel/literals/EIntegerBaseTest.java
```diff
@@ -0,0 +1,37 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+public class EIntegerBaseTest
+{
+
+  static void checkFormat (String expected, String source, String format, boolean allowLeadingSep)
+  {
+    StringBuilder sb = new StringBuilder ();
+    EIntegerBase.addSep (source, format, allowLeadingSep, 0, 0, sb);
```
phax, line 12: You could extend the tests to cover the two "0" parameters as well

jcodemodel/src/main/java/com/helger/jcodemodel/literals/AIntegerRepresented.java
glelouet: you mean add `@NonNull` on the return ? Sure. Can you do it, or I do it ?

jcodemodel/src/main/java/com/helger/jcodemodel/literals/AIntegerRepresented.java
glelouet: no, representation is `@NonNull`

jcodemodel/src/main/java/com/helger/jcodemodel/literals/AIntegerRepresented.java
phax: If it is too cumbersome for you, let me know

jcodemodel/src/main/java/com/helger/jcodemodel/literals/AIntegerRepresented.java
phax: Then I'd prefer a "if (null) throw" with the parameter - to make sure the contract of the method is clear

jcodemodel/src/main/java/com/helger/jcodemodel/literals/EIntegerBase.java
glelouet: hmm.
At first bases had different behvior but now this is split enough that it can be done, you're right.

glelouet: sorry I did not mean to start a review on the previous comment.

jcodemodel/src/main/java/com/helger/jcodemodel/literals/EIntegerBase.java
glelouet: I don't get why, but okay

jcodemodel/src/main/java/com/helger/jcodemodel/literals/EIntegerBase.java
glelouet: just to know when reading what is done in what case, it's easier to have the list of conditions that lead to a part of code.

jcodemodel/src/main/java/com/helger/jcodemodel/literals/EIntegerBase.java
glelouet: That's the auto formatter ? I believe.

jcodemodel/src/main/java/com/helger/jcodemodel/literals/IntegerRepresentation.java
glelouet: yes but then no more immutable so the proposed defaults representations could be modified.

jcodemodel/src/main/java/com/helger/jcodemodel/literals/IntegerRepresentation.java
glelouet: I just don't think it's an optimization worth the additional code, so bugs.
my b you're right it calls a new()
I do it.

jcodemodel/src/test/java/com/helger/jcodemodel/literals/EIntegerBaseTest.java
glelouet: Waht ?

jcodemodel/src/main/java/com/helger/jcodemodel/literals/EIntegerBase.java
glelouet: okay I just tried, but there are a lot of call to base values/methods so not sure (just moved the int calls, long remains)

```java
  public String format (int i)
  {
    StringBuilder sb = new StringBuilder ();
    boolean neg = i < 0;
    int posI = neg ? -i : i;
    if (neg)
      sb.append ('-');
    else
      if (positiveSign ())
        sb.append ('+');
    sb.append (prefixUpper () ? base.prefixUpperCased : base.prefixLowerCased);
    EIntegerBase.addSep (base.padBody (base.caseBody (base.intFormat.apply (posI), bodyUpper ()), padding ()),
                         separateFormat (),
                         base.allowBodyLeadingSep,
                         separateEvery (),
                         separatorSize (),
                         sb);
    return sb.toString ();
  }

  public String format (long l)
  {
    return base.represent (l, new StringBuilder (), this).toString ();
  }
```

That's because the padding is disabled for decimal base, and the case is only enabled for hexa.

jcodemodel/src/main/java/com/helger/jcodemodel/literals/AIntegerRepresented.java
glelouet: ok it's done.

glelouet: > If it is too cumbersome for you, let me know

I was not sure what this actually entailed. 
I changed to also return nonnull for self() and calling methods (in addition to representation() ), not sure if that was what you meant

jcodemodel/src/main/java/com/helger/jcodemodel/literals/IntegerRepresentation.java
glelouet: There should be java withers in the JEP that allow to alter several record fields at once with with({}) . In the meanwhile …

jcodemodel/src/main/java/com/helger/jcodemodel/literals/AIntegerRepresented.java
glelouet: The issue with not accepting null setters is that when you don't care about it, you still have to care about it.
If you're loading a settings and inject it, if your setting is missing the value to set you need to make that check on every place that can inject if, and make another dev choice to fail or not at this part.

Here we accept null pram as "I don't care", as the underlying value can't accept null. This alleviates the need to encompass all possible meanings at one time for the (dev) library user.

Now I agree that in that very specific case it's very unlikely to happen since those are in the code, not in the formatterSettings. As in, the devs are encouraged to create their own for their library.

So if you think it would create issues, go for it. There *should* not be any tests that this would make fall.

jcodemodel/src/main/java/com/helger/jcodemodel/literals/EIntegerBase.java
phax: @glelouet any comment on this one? Objections again moving it?

glelouet: @phax 

I commented though?

I even tried it and gave you what it would take, in terms of code. 

I did not submit it because it means to have the behavior in the global representation rather than in the base. which is quite dubious in term of use, since the EIntegerBase is what decides how to represent, while the IntegerRepresentation is just a series of params.

Maybe the call could be shortcircuited in the caller to direct to the EIntegerBase ?

jcodemodel/src/test/java/com/helger/jcodemodel/literals/EIntegerBaseTest.java
glelouet: @phax  I did not get what you meant here. Should I add more tests ? If yes, which ones ?

PR #184, opened 2026-08-29
"generate source" mojo iterates over the files if source is a dir.

glelouet:
The plugin can now be configured with a directory rather than an url / single file

In the example module, the pom references the `concreteYaml` directory to generate some trivial concrete classes, before the files in that directory were in a single one.

```xml
            <configuration>
              <source>concreteYaml</source> <!-- << here -->
              <rootPackage>com.helger.jcodemodel.examples.plugin.yaml.concrete</rootPackage>
```

This is done at the plugin level so generators should not be changed.

When a dir is found, all the files in it are applied to the generator, on the same JCM which is exported at the end. Therefore, each file must be valid on their own, and should not reference type from another one.

Also more docs for api.

No change is applied otherwise, and it only *allows* to use directory, so previous behavor did not change

### DIFF
--- a/plugin/plugin/src/main/java/com/helger/jcodemodel/plugin/maven/GenerateSourceMojo.java
+++ b/plugin/plugin/src/main/java/com/helger/jcodemodel/plugin/maven/GenerateSourceMojo.java
@@ -16,11 +16,14 @@ package com.helger.jcodemodel.plugin.maven;
 
 import java.io.File;
 import java.io.FileInputStream;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
+import java.util.Comparator;
 import java.util.Map;
+import java.util.stream.Stream;
 
 import org.apache.maven.plugin.AbstractMojo;
 import org.apache.maven.plugin.MojoExecutionException;
@@ -141,13 +144,24 @@ public class GenerateSourceMojo extends AbstractMojo
     {
       getLog ().warn ("discarding source param " + m_sSource + " as data is already set");
     }
-    try (final InputStream aIS = StringHelper.isEmpty (m_sData) ? findSource ()
-                                                                : new NonBlockingByteArrayInputStream (m_sData.getBytes (StandardCharsets.UTF_8)))
+    Stream <? extends InputStream> sis = (StringHelper.isEmpty (m_sData)) ? findSource ()
+                                                                          : Stream.of (new NonBlockingByteArrayInputStream (m_sData.getBytes (StandardCharsets.UTF_8)));
+    for (InputStream is : sis.toList ())
     {
-      cmb.build (cm, aIS);
+      try (is)
+      {
+        cmb.build (cm, is);
+      }
+      catch (JCodeModelException | IOException e)
+      {
+        throw new MojoFailureException (e);
+      }
+    }
+    try
+    {
       new JCMWriter (cm).setJavaFeature (findJavaFeature ()).build (dir, (IProgressTracker) null);
     }
-    catch (JCodeModelException | IOException e)
+    catch (IOException e)
     {
       throw new MojoFailureException (e);
     }
@@ -177,9 +191,10 @@ public class GenerateSourceMojo extends AbstractMojo
     if (sGeneratorClass == null)
       sGeneratorClass = findGeneratorClass ();
 
-    return StringHelper.isEmpty (sGeneratorClass) ? null : (ICodeModelBuilder) Class.forName (sGeneratorClass)
-                                                                                    .getDeclaredConstructor ()
-                                                                                    .newInstance ();
+    return StringHelper.isEmpty (sGeneratorClass) ? null
+                                                  : (ICodeModelBuilder) Class.forName (sGeneratorClass)
+                                                                             .getDeclaredConstructor ()
+                                                                             .newInstance ();
   }
 
   @Nullable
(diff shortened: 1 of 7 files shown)


glelouet:
Note : this modification is because I want to extract the implicit structure from a yaml file into a flatStructure, then from it into corresponding java class.
so something like

```yaml
a:
  b: c
  d: 5
```
would produce
```yaml
A : 
 b : String
 d : int
Root:
 a : A
```

needs a lot of time to work on it though. The goal is to generate yaml parser from yaml file. And later spring entity from yaml files :D

plugin/plugin/src/main/java/com/helger/jcodemodel/plugin/maven/GenerateSourceMojo.java
```diff
@@ -198,37 +213,85 @@ protected String findGeneratorClass () throws IOException
...
+    if (rootFile.isFile ())
+    {
+      try
+      {
+        return Stream.of (new FileInputStream (rootFile));
+      }
+      catch (FileNotFoundException e)
+      {
+        // should never happens since isFile checks for existence
+        throw new RuntimeException (e);
```
phax, line 286: Never throw RuntimeException :) Use a `new IllegalArgumentException ("Failed open file '"+rootFile.getAbsolutePath()+"'", e);` instead

plugin/plugin/src/main/java/com/helger/jcodemodel/plugin/maven/GenerateSourceMojo.java
```diff
@@ -198,37 +213,85 @@ protected String findGeneratorClass () throws IOException
...
+  /**
+   * recursively stream the files inside the file.
+   * 
+   * @param rootFile
+   *        file to browse
+   * @return a single inputstream of the file if it is a normal file. inputstreams of its children
+   *         if it is a dir. Otherwise, return empty stream.
+   */
+  @NonNull
+  protected Stream <InputStream> streamFiles (File rootFile)
```
phax, line 275: We should change the return type to a "pair (InputStream, filename)" so that error messages from the caller can provide more details

plugin/plugin/src/main/java/com/helger/jcodemodel/plugin/maven/GenerateSourceMojo.java
```diff
@@ -198,37 +213,85 @@ protected String findGeneratorClass () throws IOException
...
       final URL aURL = new URL (m_sSource);
-      return aURL.openStream ();
+      return Stream.of (aURL.openStream ());
     }
     catch (final IOException e)
     {
-      getLog ().info ("while trying to open " + m_sSource + " as a url", e);
+      getLog ().error ("while trying to open " + m_sSource + " as a file", fileException);
+      getLog ().error ("while trying to open " + m_sSource + " as a url", e);
     }
```
phax, line 261: The log errors should be split into the appropriate branches

plugin/plugin/src/main/java/com/helger/jcodemodel/plugin/maven/GenerateSourceMojo.java
glelouet: As written in the comment, this should be guarded against by the if. This is something that is really a problem if happening.

plugin/plugin/src/main/java/com/helger/jcodemodel/plugin/maven/GenerateSourceMojo.java
glelouet: yeah, I was thinking about that. The issue is that eg for urls there is no such a file name to be provided, and when the data is set, it's priority is higher and does not have a file name either.

Was still thinking about it.

plugin/plugin/src/main/java/com/helger/jcodemodel/plugin/maven/GenerateSourceMojo.java
glelouet: No, I did it exactly like this on purpose.
If the first call fails, but not the second, it means there was no point complaining about the file not existing.

glelouet:
@phax  the new sourcesFilte param : 
I named it m_sSourcesFilter first but I had errors when compiling the examples so I renamed it to sourcesFilter and it passed.

plugin/plugin/src/main/java/com/helger/jcodemodel/plugin/maven/GenerateSourceMojo.java
glelouet: to be more precise : if there is an error here, we MUST change the code, it should never happen whatever the user sets.

phax: But never throw RuntimeException - its just bad practise. If you code the exception, give it a valid message otherwise if it happens, at least we know what it is. Use IllegalStateException if you prefer

plugin/plugin/src/main/java/com/helger/jcodemodel/plugin/maven/GenerateSourceMojo.java
```diff
@@ -288,6 +293,9 @@ protected Stream <InputStream> streamFiles (File rootFile)
       }
       else // file is directory
         return Stream.of (rootFile.listFiles ())
+                     .filter (f -> sourcesFilter == null ||
+                       sourcesFilter.isBlank () ||
+                       f.getName ().toLowerCase (Locale.getDefault ()).contains (sourcesFilter.toLowerCase ()))
```
phax: Add `Locale.ROOT` to the second `toLowerCase` as well

glelouet: haaaaaaa I was so proud to have found the first one :P

plugin/plugin/src/main/java/com/helger/jcodemodel/plugin/maven/GenerateSourceMojo.java
glelouet: There is no valid message if this happens.
It's the same as 1+1=3.
This is not illegalState : this is the jvm checking for a file existing, then proceeding as the file exists, and the file does not exist when opening it. 
Illegalstate is when the calls have lead to an illegal state.

A valid exception in that case could be an Error as it indicates critical issue that goes beyond what the user is expected to do / deal with.
What about java.io.IOError ?

phax: Well there is always an opportunity, especially when files are involved, that something "unexpected" can happen. IOException does not really match (and is a checked exception - so better UncheckedIOException). I still prefer IllegalState because the state of the application is "invalid"...

PR #186, opened 2026-08-30
keep inputstream source when generating from the plugin

glelouet:
Instead of getting a stream of inputstream, we also get their source information, if any. 

That source information is transmitted down to the generators, so they can use it, eg to deduce class name based on the file.

It is also printed on error, to know which file / url generated the issue.

### DIFF
--- a/plugin/generators/csv/src/main/java/com/helger/jcodemodel/plugin/generators/csv/CSVGenerator.java
+++ b/plugin/generators/csv/src/main/java/com/helger/jcodemodel/plugin/generators/csv/CSVGenerator.java
@@ -15,7 +15,6 @@
 package com.helger.jcodemodel.plugin.generators.csv;
 
 import java.io.BufferedReader;
-import java.io.InputStream;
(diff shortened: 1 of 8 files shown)


glelouet:
sorry, should be draft, I realize the pair is not transmitted to the generator which could use that information

glelouet:
I made a PR rather than a discussion for this one. The reason is that I was thinking how to do that before, and how it could help the generators.

phax:
I have the concept of `IReadableResource` in my ph-io base library - with `ClassPathResource`, `FileResource` and `URLResource` with the sanity methods for `InputStream` and `Reader` - please have a look

glelouet:
Can't find the repo. Can you link @phax ?
Or better, change the PR to incorporate your idea instead.

phax:
@glelouet start from https://github.com/phax/ph-commons/blob/master/ph-io/src/main/java/com/helger/io/resource/ClassPathResource.java

PR #188, opened 2026-08-30
JAtom int/long direct call to representation

glelouet:
Basically the IntegerRepresentation is only the params. The EIntegerBase is the one doing the formatting.

also renamed "represent()" to "format()", more explicit

also renamed field representation to m_aRepresentation, using getter representation() instead.

### DIFF
--- a/jcodemodel/src/main/java/com/helger/jcodemodel/literals/IntegerRepresentation.java
+++ b/jcodemodel/src/main/java/com/helger/jcodemodel/literals/IntegerRepresentation.java
@@ -199,18 +199,4 @@ public record IntegerRepresentation (
     return suffixUpper == this.suffixUpper ? this : with (ir -> { ir.suffixUpper = suffixUpper; });
   }
 
-  //
-  // actual formatting is delegated to the base
-  //
-
-  public String format (int i)
-  {
-    return base.represent (i, new StringBuilder (), this).toString ();
-  }
-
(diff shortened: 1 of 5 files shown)


jcodemodel/src/main/java/com/helger/jcodemodel/literals/IntegerRepresentation.java
```diff
@@ -199,18 +199,4 @@ public IntegerRepresentation sufixUpper (boolean suffixUpper)
     return suffixUpper == this.suffixUpper ? this : with (ir -> { ir.suffixUpper = suffixUpper; });
   }
 
-  //
-  // actual formatting is delegated to the base
-  //
-
-  public String format (int i)
```
phax, line 206: Why remove these sanity methods? Isn't that something you usually like?

glelouet: I just moved the call ? instead of applying to the Representation, which calls the base with itself as params, I directly call the base with the Representation as param.

This was something you mentionned in the previous PR, but was left off when merging.

https://github.com/phax/jcodemodel/pull/182/changes/BASE..3bdb5451ab1ca63c54468698a14c0049dc0a98b6#diff-27921528c84e44ffe0fa1c30100ce28fa45cbec7042924072abd01196e7eb229R64

PR #189, opened 2026-09-05
added explanation of self generator, from my experiences and failures

glelouet:


phax:
Here is some feedback, with a little help on the details:

* The `<optional>true</optional>` dep on generator isn't needed. The doc says it's what sequences the modules — it isn't. Maven's reactor sorter already accounts for plugin `<dependencies>`, and `generated` declares the `generator` there.

Markdown bugs

- docs/extending.md:92 — `* generated*` renders literally (space kills the emphasis). Should be `*generated*`.
- :145-184 — the ordered list 1.–3. is terminated by the unindented code fence, then 4. starts a fresh list. Indent the fences 3 spaces to keep them inside the item.
- Step 3's snippet is a complete `<build>` block, step 4's is a bare `<plugin>`. Pasting both yields two `<build>` sections. Merge them, or say "into the same <plugins> list".
- Empty `<configuration></configuration> `in the plugin snippet — drop it.
- `<id>generate source</id>` — a space in an execution id breaks `mvn ...:generate-source@<id>`. Use `generate-source`.
- Still no newline at EOF (the PR rewrites the last line — good chance to fix).
- Trailing whitespace on 13 lines; lines 93 and 140 contain only a space.

Wording

- Opening sentence is garbled: "JCM allows to define and use code generators to call a plugin by maven, that writes code generated by the generators." Reads as generators calling the plugin. Suggest: "JCM ships a Maven plugin that runs a code generator at build time and writes the resulting sources into your project."
- :43 — "create a new module in the the plugin/generators directory" (PR edits this line; typo survives).
- :37 — "This part focus" --> "focuses".
- Space-before-punctuation throughout (;, :,  !) — French convention. Some was already there; worth unifying rather than spreading.
- First person: "I assume you want..." and "In my case I add the plugin call in a generate profile". docs/contributing.md uses "my" as Philip's voice, so a contributor's "I" is ambiguous here. Either make it impersonal, or make the attribution explicit.
- :50 — "...nothing changes too much : " dangles a colon straight into a heading.

glelouet:
> Step 3's snippet is a complete <build> block, step 4's is a bare <plugin>. Pasting both yields two <build> sections. Merge them, or say "into the same list".

They are split because there are two different things : step 3 is configuring the other modules to work with the generated source folder, and step 4 is configuring the jcm plugin to be launched with the generator

glelouet:
> The <optional>true</optional> dep on generator isn't needed. The doc says it's what sequences the modules — it isn't. Maven's reactor sorter already accounts for plugin <dependencies>, and generated declares the generator there.

I remind having issues linked to that. 

Even though indeed it is supposed to work this way
https://maven.apache.org/guides/mini/guide-multiple-modules.html#Reactor_Sorting
> a plugin dependency on another module in the build

IIRc it was when a version was already in cache. 

but okay we can remove that one

glelouet:
> :145-184 — the ordered list 1.–3. is terminated by the unindented code fence, then 4. starts a fresh list. Indent the fences 3 spaces to keep them inside the item.

I tried, did not work, I give up. Had other things to do rather than fighting with syntax. Feel free to fix it, if you know how - personally I already spent a few hours on that PR, including making the errors again to have the results, and trying to compile with JCM 4.3.1-snapshot because I had not copypasted the version.

Also the eclipse md parser may be slightly different from the github one. I never know if the result will be okay.

edit : used unordoned list , no more numbers.

glelouet:
> <id>generate source</id> — a space in an execution id breaks mvn ...:generate-source@<id>. Use generate-source.

it does not break for me ?
> [INFO] --- jcodemodel:a.b.c:generate-source (generate source) @ generated ---

glelouet:
> :43 — "create a new module in the the plugin/generators directory" (PR edits this line; typo survives).

I don't see the typo :P 

edit : finally saw the the typo, fixed.

glelouet:
> Space-before-punctuation throughout (;, :, !) — French convention. Some was already there; worth unifying rather than spreading.

Yeah it's french requirement so even if I write an english doc I still follow it :P 

I let you fix them, I just don't see them and would even "french fix" ones that do not follow it.

(In french typography, two-shape punctuation symbols must follow a space and be followed by one)

glelouet:
> :50 — "...nothing changes too much : " dangles a colon straight into a heading.

Yeah, it's the link to the next part. I changed to a period instead.

glelouet:
@phax I think I either discussed or added each of your point.

When the PR is accepted, don't hesitate to change what is weirdly worded : when I add things, I want to be sure the important data is present, not that this can become a best-seller :-D

phax:
Thanks a million for your efforts

glelouet:
As I said, I was working on making my own generator on another project (for the record, to make an openapi parser with even-driven approach and maybe spring representation) and realized it took more time to implement  than I was expecting it.

Even had to juggle with the java version but in the end it did not matter (so not present in the doc).

Using SNAPSHOT revealed that the fix I implemented (the one where the file error is only printed when URL also fails) works(as without it the URL resource is not even loaded), even though when calling an url and producing an error (in the generator) it also prints the file error which is quite confusing.

PR #190, opened 2026-09-06
maven plugin now concats data and source params streams.

glelouet:
STATUS : stale
A potential issue with using stream of URL arose, a security check needs to be performed

---

Previously, in the plugin params, if "data" was set then "source" was ignored ; now they are both used, providing several inputstreams to apply the generator against.

If none are present, the plugin still resolves a Stream.of(null) : this way a generator which does not need data would still work (is called against a null inputstream), while a generator that requires data would fail (throwing NPE) rather than be silently skipped.

Also fixes the plugin not trying to fetch the source as an url, and showing errors when no error happened.

Also fixes the plugin silently skipping when the source could not be loaded.

This was possible after the patch listing the children when the source is a directory, as this patch enabled having several sources.

### DIFF
--- a/plugin/generators/csv/src/main/java/com/helger/jcodemodel/plugin/generators/csv/CSVGenerator.java
+++ b/plugin/generators/csv/src/main/java/com/helger/jcodemodel/plugin/generators/csv/CSVGenerator.java
@@ -15,7 +15,6 @@
 package com.helger.jcodemodel.plugin.generators.csv;
 
 import java.io.BufferedReader;
-import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.util.Map;
 import java.util.stream.Stream;
@@ -24,6 +23,7 @@ import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.Nullable;
 
 import com.helger.base.string.StringHelper;
+import com.helger.jcodemodel.plugin.maven.ISourcedInputStream;
 import com.helger.jcodemodel.plugin.maven.generators.AbstractFlatStructureGenerator;
 import com.helger.jcodemodel.plugin.maven.generators.JCMGen;
 import com.helger.jcodemodel.plugin.maven.generators.flatstruct.FieldOptions;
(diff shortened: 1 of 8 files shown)


phax:
This piece of code should afterwards be assessed from a security perspective - besides that it looks good for me

phax (approved):

glelouet:
> This piece of code should afterwards be assessed from a security perspective - besides that it looks good for me

waht ? I don't get it.

glelouet:
@phax  I made a few changes to merge the other PR correctly. Please verify again.

phax:
Because we work with streams of URLs we need to see if this can be used to get in remote data we don't want to

glelouet:
> Because we work with streams of URLs we need to see if this can be used to get in remote data we don't want to

No ?

We work with SourcedInputStream, one of which at most is created from an url.

```xml
<data>my raw data, to anger Gordon Ramsay</data>
<source>http://my.source.txt</source>
```
can only result in one inputstream based on the raw data, and one inputstream based on the url.

Maybe it's still an issue, you're right. But the count of URL-based stream is (0,1)

glelouet:
@phax should I return a list of SourcedInputStream instead ? In the present calls, the stream is used with .tolist() so that would barely change anything, while not having a stream containing indirect url(s) .

phax:
No idea - I would need to look at it once it is merged - but I currently have a different focus

glelouet:
Okay.

Since it's a potential security issue, I rather draft it, and wait until you find the time to be sure there is no added issue.

phax:
Yes please draft it - I will merge in a temp branch and check what I can do with it ;-)

glelouet:
Already did.

I also re requested a review. Does not mean you have to do it now :)

I update the first post. I already updated it to present the improvement in a way that can be easily integrated (I believe) during the version release.

phax:
Please mark the PR as ready

glelouet:
@phax  since this PR is a minor, the next version should be 4.4.0 instead of 4.3.1.
The fact that we allow to load several files from directories is IMO a feature.

PR #191, opened 2026-09-07
maven plugin now concats data and source params streams. (#190)

phax:
* maven plugin now concats data and source params streams.

Previously, "data" was superseeding the "source" ; now they are both used.

If none are present, still returns a Stream.of(null) so that a generator which does not need data would still work (is called against a null inputstream), while a generator that requires data would fail (with NPE) rather than be silently skipped.

* keep inputstream source when generating from the plugin (#186)

* plugin sources are kept with the inputstream

Now the source responsible is returned in the error

* sourcedinputstream transmitted to the generators.

* activate specific projects formatter and reformat them :P

---------

* removed the default build(JCM)

since the plugin should never had no data to transmit, no point in this method that was not called to start with.

---------

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,20 @@ Add the following to your pom.xml to use this artifact (where `x.y.z` denotes th
 
 # News and noteworthy
 
+v4.3.1 - work in progress
+* Maven plugin: the `data` and the `source` parameters are now both applied, instead of `data` superseeding `source`. The generators receive the new `ISourcedInputStream`, that keeps the source an inputstream was created from. See [#186](https://github.com/phax/jcodemodel/issues/186) and [#190](https://github.com/phax/jcodemodel/pull/190) - thx @glelouet
+* Maven plugin: the source files are opened one at a time and always closed, even if the generation fails
+* Maven plugin: directories are no longer skipped by `sourcesFilter` - the filter applies to the files only, as documented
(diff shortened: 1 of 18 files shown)


glelouet:
@phax  the SourcedInputStream should be @nonNull instead. The internal inputstream can be null, but only for the NullSourcedInputStream - thats why I changed the code

phax:
Done in master

glelouet:
@phax  okay, I was thinking maybe you changed the business.

phax:
@glelouet  I don't see your comment on the merge commit readme.
No worries, this will not stay like this - many more changes of course

glelouet:
@phax  I was saying you missed the two bug fixes regarding not fetching the url, and silently not calling the generator when the directory specified as source did not contain valid file.

see first message of
https://github.com/phax/jcodemodel/pull/190

> Also fixes the plugin not trying to fetch the source as an url, and showing errors when no error happened.
> 
> Also fixes the plugin silently skipping when the source could not be loaded.

The idea is that now, the list of SourcedInputStream is never empty : if no correct inputstream is selected, then still returns a single nullsourcedinputstream .

Its then up to the generator to handle the nullsourcedinputstream. Typically a generator that does not use it (ie helloworldgenerator)  would ignore it ; while a generator that needs it (each other so far) would fail as it can't work against no source.

PR #193, opened 2026-09-08
fix null inputstream behavior

glelouet:
- ensures the generator receives at least one SourcedInputStream (NULL if none actually transmitted)
 - helloworldgenerator prints a warning when it receives a non null source (as if several are received, it will apply several times to the JCM with the same params and this leads to error when creating the same class twice). The example has a data set to show this works :) 
 - csv and json generators print a warning if the inputstream is null. They are designed to work against a source, so not receiving a source to work against may be an issue.

### DIFF
--- a/examples/plugins/helloworld/pom.xml
+++ b/examples/plugins/helloworld/pom.xml
@@ -64,6 +64,7 @@ limitations under the License.
             <configuration>
               <generator>com.helger.jcodemodel.plugin.generators.helloworld.HelloWorldGenerator2</generator>
               <outputDir>src/generated/java2</outputDir>
+              <data>useless data</data>
               <params>
                 <name>Hello2</name>
                 <value>world2</value>
--- a/plugin/generators/csv/src/main/java/com/helger/jcodemodel/plugin/generators/csv/CSVGenerator.java
+++ b/plugin/generators/csv/src/main/java/com/helger/jcodemodel/plugin/generators/csv/CSVGenerator.java
@@ -64,7 +64,10 @@ public class CSVGenerator extends AbstractFlatStructureGenerator
   {
     final InputStream is = source.inputStream ();
     if (is == null)
-      return Stream.empty ();
+      throw new IllegalArgumentException ("generator " +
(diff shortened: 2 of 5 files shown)


glelouet:
it fails because the tests assume receiving a null source is correct for csv ; it's not.

> 2026-09-08T00:16:34.3983679Z [ERROR] com.helger.jcodemodel.plugin.generators.csv.CSVGeneratorTest.testNoSource -- Time elapsed: 0.018 s <<< ERROR!
> 2026-09-08T00:16:34.3985115Z java.lang.IllegalArgumentException: generator CSVGenerator needs to receive inputstream, received NullSourced[]
> 2026-09-08T00:16:34.3986233Z 	at com.helger.jcodemodel.plugin.generators.csv.CSVGenerator.loadSource(CSVGenerator.java:68)
> 2026-09-08T00:16:34.3987637Z 	at com.helger.jcodemodel.plugin.generators.csv.CSVGeneratorTest.testNoSource(CSVGeneratorTest.java:103)

glelouet:
The issue with accepting null source for generators that need one is that it is not clear why it stopped producing classes : was the plugin disabled ?

Maybe we can warn instead of illegalargumentexception

glelouet:
Now it warns in the tests.

[main] WARN com.helger.jcodemodel.plugin.generators.csv.CSVGenerator - generator CSVGenerator did not receive a valid source, received NullSourced[] instead

and again, the helloworld warns too

[WARNING] generator HelloWorldGenerator2 received a source DirectSourced[inputStream=com.helger.base.io.nonblocking.NonBlockingByteArrayInputStream@4cbb217e] which is discarded. If multiple sources are provided, this may lead to runtime error

phax:
Sounds like a plan :)

glelouet:
@phax  also you may have noticed I don't appear twice anymore on the commits. I changed my mail address on my local machine to the noreply of git, instead of n@mo .

phax:
Okay cool - honestly I don't care about that one

PR #194, opened 2026-09-08
Changing the doc of the plugin + prefix "jcm" + fix mojo not testing for blank

glelouet:
1. The prefix "jcm" allows to call the plugin with jcm:generate-source. This just shorter than previous "jcodemodel" which was deduced from the artifactId

2. The mojo javadoc has been reworked for all the param and main class with simple sentences that are more easy to read, especially since the help does not show formatting correctly. No more md doc for those, no more html. Details are moved to other methods.

3. bug fix : the mojo params were tested for StringHelper.isEmpty ; they are now tested for blankness instead of emptyness.

after this, the result of `mvn -q -T2C -ntp install && mvn help:describe -Dplugin=jcm -Ddetail` is :

```
Name: MVN JCodeModel Maven Plugin
Description: Maven plugin to call jcodemodel.
Group Id: com.helger.jcodemodel
Artifact Id: jcodemodel-maven-plugin
Version: 4.3.1-SNAPSHOT
Goal Prefix: jcm

This plugin has 1 goal:

jcm:generate-source
  Description: Apply a java source generator per source provided. The
    generator is required, as a dependency, and/or as a configuration.
  Implementation: com.helger.jcodemodel.plugin.maven.GenerateSourceMojo
  Language: java
  Bound to phase: generate-sources

  Available parameters:

    classHeader
      User property: jcodemodel.classheader
      Documentation added to the generated unit classes header,
typically a
      license. Only the main class of a unit is impacted, and only if
the
      generator did not already set the header.

    data
      User property: jcodemodel.data
      When this param is set to non-blank, the value is used as an
additional
      source transmitted to the generator.

    generator
      User property: jcodemodel.generator
      The fullly qualified name of the generator used. Only needed if
you don't
      have unique standard generator in the dependencies, or you want to
      specify a non-default one.

    javaFeature
      User property: jcodemodel.java.feature
      Java feature (major release version) the generated class files are
      targeted at. When unset the default of
      com.helger.jcodemodel.writer.JCMWriter.DEFAULT_JAVA_FEATURE is
used.

    outputDir (Default: src/generated/java)
      User property: jcodemodel.outdir
      target directory to place the generated java files into. The
directory is
      created but not cleaned.

    params
      User property: jcodemodel.params
      direct Map of params to transmit to the generator. The generator
used
      will be configured using this map, deciding which set params it
should
      use.

    rootPackage
      User property: jcodemodel.rootpackage
      fully qualified package to create the classes in. Example
      ${project.groupId}.${project.artifactId}

    source
      User property: jcodemodel.source
      source of the data to transmit to the generator when building the
model.
      Can be a url, a file, a directory.

    sourcesFilter
      User property: jcodemodel.sourcesFilter
      If the source is a directory, and this param is not blank, only
the files
      in that directory with a last name containing this param (ignoring
case)
      will be passed to the generator. Typically used with file
extension, eg
      ".txt"
```

Note : I tried changing the plugins version to handle md javadoc and correct printing, did not work.

### DIFF
--- a/plugin/plugin/src/main/java/com/helger/jcodemodel/plugin/maven/GenerateSourceMojo.java
+++ b/plugin/plugin/src/main/java/com/helger/jcodemodel/plugin/maven/GenerateSourceMojo.java
@@ -42,7 +42,6 @@ import org.apache.maven.project.MavenProject;
 import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.Nullable;
 
-import com.helger.base.string.StringHelper;
 import com.helger.jcodemodel.JCodeModel;
 import com.helger.jcodemodel.exceptions.JCodeModelException;
 import com.helger.jcodemodel.plugin.maven.ISourcedInputStream.DirectSourced;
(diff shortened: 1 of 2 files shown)


plugin/plugin/src/main/java/com/helger/jcodemodel/plugin/maven/GenerateSourceMojo.java
```diff
@@ -51,28 +50,24 @@
...
-/// 
+/**
+ * Apply a java source generator per source provided.<br />
+ * The generator is required, as a dependency, and/or as a configuration.
+ */
 @Mojo (name = "generate-source", threadSafe = true, defaultPhase = LifecyclePhase.GENERATE_SOURCES)
 public class GenerateSourceMojo extends AbstractMojo
 {
 
+  static boolean isBlank (String s)
```
phax, line 61: You can use `StringHelper.isEmptyAfterTrim`

plugin/plugin/src/main/java/com/helger/jcodemodel/plugin/maven/GenerateSourceMojo.java
```diff
@@ -51,28 +50,24 @@
...
+  {
+    return s == null || s.isBlank ();
+  }
+
+  static boolean isNotBlank (String s)
```
phax, line 66: and `StringHelper.isNotEmptyAfterTrim`

phax (approved):

phax:
Are we ready for a 4.3.1 release?

plugin/plugin/src/main/java/com/helger/jcodemodel/plugin/maven/GenerateSourceMojo.java
glelouet: I checked and it uses trim() which does not discard tabs :D

plugin/plugin/src/main/java/com/helger/jcodemodel/plugin/maven/GenerateSourceMojo.java
glelouet: same :D

glelouet:
@phax tomorrow, I should have finished the last pending PR :)

glelouet:
@phax feel free to use StringHelper instead. I considered using AfterTrim but trim() does not seem to remove tabs, only true whitespace. If I'm wrong then use StringHelper instead.

phax:
No worries - I merged and extended ph-commons with a "is(Not)Blank" for the next version. No need right now

glelouet:
@phax  also there is https://github.com/phax/jcodemodel/pull/193

PR #195, opened 2026-09-10
Less paranthesis

phax:
As the follow-up on #183


phax:
@glelouet as usual, I don't see your comments on the commit https://github.com/phax/jcodemodel/commit/25f328d670aecde21b41e87a4e8097c20996599c

Please repost here

glelouet:
@phax  that's annoying.

There was several comments

1. you gave precedence 15 to assignment and lambda which is incorrect : `l = a -> f=a` is `l = a -> (f=a)`. This happened because you replace the call to ordinal() with an internal field that is the ordinal … please revert to ordinal.

2. in the generated test classes, with the always settings, some parenthesis disappeared. They should not.

phax:
@glelouet please recheck master

glelouet:
@phax  lgtm 

Though What I said was actually wrong, I need to make a few tests.
` Consumer<Object> co = o->myField=o ;` has two assignement, yet the actual parenthesized would be `co=(o->(myfield=o)) ;`

phax:
So I am on vacation and most likely can't do a release built in-between....

glelouet:
That gives me some time for the next PR … :evil_face:

phax:
Evil face does not render on GitHub - sorry ;-)

phax:
Btw for that feature I really recommend test driven development - there are so many permutations

glelouet:
nah, it's actually just a bit of code. But please give your advice in the corresponding PR :-D  :very_evil_face:

phax:
@glelouet do you intent to do anything additional on this matter?

glelouet:
@phax ? No ?

phax:
good :)

PR #196, opened 2026-09-10
TypedExpressions

glelouet:
Summary for the release:
 - add TypedExpression hierarchy and links to provide explicit method and field access in common classes. Limited to String, Object, Arrays and Numerics so far.

# Goal and constraints

The goal is to have more control over the types in the expressions : when at design time, a dev knows the type of the variables manipulated at compile and run time.

Knowing the underlying type is beneficial in two ways: 
 1. propose what we can do : if we know we have a String, then we can call length() on that String. We can then write`myIntegerEx = myStringEx.length()` 
 2. don't propose methods that don't make sense. StringExpression for example does not propose a mult method, but it proposes a plus method.

## Target usage

I want to be able to write things like 
```java
// new
myMethod.body().add(myStringExpr.charAt(myIntExpression));

// old
myMethod.body().add(myIJExpr.invoke("charAt").param(myIJExpression2)) ;
```

This would have design time check (by compiling the design call)

## Constraints

This means a specific wrapper class for each java base class we want to embed.

The primitive classes must be done manually, but then the java.lang classes can be done automatically by JCM, with a selection of the ones performed for incremental tests. 

The templated classes would be ignored at first,. The compiler will require additional work later.

# Implementation

### Added classes

 - The base Interface is `com.helger.jcodemodel.expressions.ITypedExpression` with only an `IJExpression raw()` . Main implementation is `com.helger.jcodemodel.expressions.typed.TypedExpressionWrapper <RunTimeType>`
 - primitve types have a dedicated hierarchy to manage the various operator and their promotions in the java language, including static tooling methods eg `IntExpression IntExpression::of(int)`, `IntExpression IntExpression::of(IJExpression untyped)`, and `IntExpression IntExpression::of(ITypedExpression other)`.
 - String has specific operators so its own StringExpression class too. Mirrors several String methods, eg `CharRepresentation StringExpression charAt( SubIntRepresentation<?,?,?> index)` translates into `that.charAt(index)`
 - Object is required by String, so also has its own handwritten implementation.
 - Void and array types are in the primitive package. They are needed for the return type : The `Object::notify` returns void, so the `ObjectExpression::notify` returns a VoidStatExpression ; the `String::getBytes` returns an array of bytes, so the `StringExpression::getBytes` returns a `ByteExpression.ByteArrExp`

### Modifications to existing classes

 - the JExpr.assign overloaded to accept a ITypedExpression instead of its right IJExpression, resolving to the later using raw().
 - same for JInvocation::arg, JBlock::_if, JFor::test and JFor::update ._if and test specificallyrequire boolean expression.
 - JMethod proposes a tooling paramTyped method, called by the concrete XExpression in their static builder, to add a param of known type. They are not present in the JMethod because there could be too many of them ; plus some are built later so it simplifies the dependencies.
 - most other usages should not be touched. I want to avoid the binding between that new hierarchy and old code.
 - JMethod proposes a tooling paramTyped method, called by the concrete XExpression in their static builder, to add a param of known type. They are not present in the JMethod because there could be too many of them ; plus some are built later so it simplifies the dependencies.

## Primitive types

The hierarchy for numerics is a bit complex, since java inheritance makes things complicated : You can't have FloatExpression extend DoubleExpression while declaring to implement ITypedExpression<Float> since that would conflict with the DoubleExpression implementing ITypedExpression<Double> 

Instead, there are abstract templated classes : the root class of numeric expressions is 
```java
ANumericExpression <
    T,
    Self extends ANumericExpression <T, Self, ?>,
    PosType extends ANumericExpression <?, ?, ?>
  > extends
    TypedExpressionWrapper <T>
```

The `DblExpression` extends it but the *usual* param translating to a double *should* be `ANumericExpression<?, ?, ?>` to allow to pass both DblExpression and CharExpression for example. 
A methods `double max (double a, double b)` *should* be mirrored as `DblExpression max( ANumericExpression<?, ?, ?> a, ANumericExpression<?, ?, ?> b)`

There are 4 abstract primitive numeric classes : for double, float, long, int. They each add overloaded numeric operators when promotion is in place : `double+numeric := double` but `float+char:=float`, in the later the char is promoted to float, so `ASubFloatExpression` has a `public FltExpression plus (ASubFloatExpression <?, ?, ?> other)` to only promote the result type up to float, while the methods inherited from ANumericExpression remains and promotes to double. 

The `ASubIntExpression` also adds
 - several bitwise operators which are only avail for integral types,
 - the "plus" operator which is an explicit int promotion for all the sub int types (byte, char, int, short) .

# Next

This only shows basic hierarchy. More complex one, including expression compiler, will be in another PR to avoid overloading the reviewer.


glelouet:
Base arithmetic done, including numeric types promotions and String operator+ .
I gave up overloading methods with primitive types as those need be overloaded for all types to avoid conversion :/ 

remains : 
1. make full String conversions, with placeHolder Wrapper<T> when the returned type is not known yet (Stream…) and //TODO comment to replace them later.
2. add the array expression. Issues with component not getting the correct type maybe.
3. start compiler to make selected java.lang class expressions in src/generated/java (collections, Stream, etc.)

phax:
This is still a draft, so I am holding my horses...

glelouet:
You can look for the global idea, but the change would come from the other PR that needs them ^^

glelouet:
@phax if you mean that you want to do a release, do it without this. I don't plan to complete before at least 1-2 week(s)

glelouet:
ok actually it seems to be working, I think any change can be done in another PR.

ready to review.

No need to release it though.

PR #198, opened 2026-09-21
Typed expression generator

glelouet:
This PR  based on https://github.com/phax/jcodemodel/pull/196 adds a generator to the plugins, that creates typedExpressions for specifically set classes.
To this point it's a new mojo, but later I will make a generator instead

----

Depends on : 
 - https://github.com/phax/jcodemodel/pull/196
 - https://github.com/phax/jcodemodel/pull/199

----

Status : 

 - [x]  test the water with simple mojo, looking at the result.
   - [ ] make it a generator instead of a mojo.
 - [x] mirror base classes for String, primitives, Object. see https://github.com/phax/jcodemodel/pull/196
 - [x] mirror public methods
   - [x] avoid Object methods
     - [x] *really* avoid Object methods, instead of hardcoding their names :P 
   - [x] avoid name clashes for methods
     - [ ] Check using parent methods too ?
   - [ ] don't duplicate method declaration (if get has same signature in parent class, don't recreate it)
   - [x] use generics for return type
   - [x] use generics for parameters
 - [ ] mirror static methods. They need be static call with JCM param in the return type (the concrete one), eg StreamExpression.of(jc, T... objects). Static methods should only be present in the final (return type) implementation.
   - [ ] mirror constructors . They are basically static methods, also only in concrete implementation.
 - [ ] mirror public instance fields.  Make them methods ? Or keep them as final fields ? Assignment can't be done directly (with = ), not sure which is better.
   - [ ] mirror public static fields. Those need a JCM to ref the linked class, so for uniformity instance fields should also be methods ?
 - [ ] use more classes to test : Stream, ByteArray.
   - [ ] specify a full package instead of listing all its classes.
 - [ ] integrate actual method parameters name, "arg0" is useless.


glelouet:
done the generics for methods return/parameters.

remains

 - static methods
 - instance fields
 - static fields
 - make it a generator instead of a mojo.
 - use more classes
 - try to use packages completely
 - find out how to integrate actual method parameters name, "arg0" is useless.

glelouet:
To be fair, this is a bit more complex than I expected at first :P 

I will do the basic so no static method, no fields, no package, only next thing is make it a generator and that will be shippable as a new feature.

PR #199, opened 2026-09-22
Adding JCM ref(Type)

glelouet:
This allows to copy existing java methods' parameterized params and ret types.

See the tests added in JCM for use example


glelouet:
@phax  this is a short one, should be quick to validate/reject

phax (approved):

